PR #27, opened 2017-04-05
Implement initial lifecycle phases for events (no services actually u…

### LINKED ISSUE #26: Add new events lifecycle methods in preparation for supporting consumers (such as Lambdas) from producers (such as S3, SNS, etc.).
See notes from notebook, this will need to be documented eventually.

dsw88:
…se it yet)

Resolves #26


dsw88:
Circle tests didn't pass because it couldn't get binaries for Node. I ran unit tests locally and they succeeded.

PR #108, opened 2017-05-10
Add support for HTTPS in Beanstalk

dsw88:
This commit switches to use the ALB in Beanstalk. It supports
HTTP and HTTPS. If you use HTTPS then you must specify a certificate
from the ACM service by ID.


dsw88:
See http://handel.readthedocs.io/en/latest/supported-services/beanstalk.html for information on using HTTPS via the 'routing' element.

PR #132, opened 2017-05-19
Add initial apiaccess service support

### LINKED ISSUE #87: How to grant the aws-accounts-to-cmdb lambda function access to call the organization.ListAccounts  API
Here is the current error:
`botocore.errorfactory.AccessDeniedException: An error occurred (AccessDeniedException) when calling the ListAccounts operation: You don't have permissions to access this resource.`

dsw88:
This service allows you to grant read-only access to AWS services
in an application. The service doesnt create any resources, it only
returns policy statements for consumption by other Handel services.

The initial service only supports the organizations service. More
can be added as needed.

Resolves #87

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+**/.DS_Store
 docs/_build
 out/
 node_modules/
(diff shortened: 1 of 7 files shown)


dsw88:
See http://handel.readthedocs.io/en/latest/supported-services/apiaccess.html.

PR #186, opened 2017-06-26
added acl parameter to s3 service

### LINKED ISSUE #167: S3 - Support Bucket-level ACLs
Currently in the S3 service you can't set any bucket-level ACLs, which some people may want to do occasionally. Bucket-level ACS are very broad policies for things like "Grant everyone in the world public read access to this bucket without authentication".

This feature should be accompanied by a big warning about what happens if you set "PublicRead" and other potentially dangerous ACLs on the bucket. We may want to not allow certain extra dangerous bucket-level ACLs like "PublicReadWrite"

ndpete:
This adds the bucket-level ACL's to the S3 service. It should be noted that by AWS specs this applies to the bucket. New objects added to the bucket need there own permissions added. For example if I create a bucket with the ACL PublicRead and then after the bucket is created I add a text file without specifying an ACL or policy the text file _will not_ inherit the PublicRead ACL, the result is a bucket that you can see the object but not access the object.

Hope that all makes sense.

resolves #167

### DIFF
--- a/docs/supported-services/s3.rst
+++ b/docs/supported-services/s3.rst
@@ -12,7 +12,6 @@ Service Limitations
 -------------------
 This service currently only provisions a bare-bones S3 bucket for data storage. It does support versioning, but the following other features are not currently supported:
 
-* Bucket-level ACLs.
 * CORS configuration
 * Bucket lifecycle
 * Bucket logging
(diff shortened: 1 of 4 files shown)


ndpete:
Needs test fixed and a test written to check for an error.

dsw88:
Sorry I was out the last two days so I'm just now looking at this, but this feature looks great!

PR #213, opened 2017-07-20
Alexa skill kit Service

### LINKED ISSUE #206: Alexa - Add the Alexa Skills Kit as an Event Producer for Lambda functions
In order for a Lambda function to be used as an Alexa Skills Kit skill one needs to add a 'trigger' to the lambda function.  Here is a screenshot of the page in the console to do it.
screen shot 2017-07-13 at 12 41 36 pm

It's not a difficult process but in order for developers to not have to remember to do manually that every time they create a new lambda function with Handel that will be an Alexa skill, I'm requesting that the Alexa Skills Kit be added as a new Handel service that can function as an event producer for Lambda functions created by Handel.

The new Alexa Skills Kit service probably wouldn't need any options.

ndpete:
Here's my first shot at the Alexa Skill Kit Service

I'm sure there is some more refinement that could be done but this is a working version.
The commit history is a little messy cause I had to merge in the IoT changes and what not but I think it makes sense.

resolves #206


lib/services/alexaskillkit/index.js
```diff
@@ -0,0 +1,90 @@
...
+    return deployPhaseCommon.deployNotRequired(ownServiceContext, SERVICE_NAME);
+}
+
+exports.consumeEvents = function (ownServiceContext, ownDeployContext, producerServiceContext, producerDeployContext) {
+    return Promise.reject(new Error(`The ${SERVICE_NAME} service doesn't consume events from other services`));
+}
+
+exports.produceEvents = function (ownServiceContext, ownDeployContext, consumerServiceContext, consumerDeployContext) {
+    return new Promise((resolve, reject) => {
+        winston.info(`${SERVICE_NAME} - Producing events from '${ownServiceContext.serviceName}' for consumer ${consumerServiceContext.serviceName}`);
```
dsw88: You'll want to return a ProduceEventsContext here, which is what the framework expects to have if it needs to use it in other places. Here's what you would do after the log statement:
```
return resolve(new ProduceEventsContext(ownServiceContext, consumerServiceContext));
```

lib/services/alexaskillkit/index.js
```diff
@@ -0,0 +1,90 @@
...
+exports.unPreDeploy = function (ownServiceContext) {
+    return deletePhasesCommon.unPreDeployNotRequired(ownServiceContext, SERVICE_NAME);
+}
+
+exports.unBind = function (ownServiceContext) {
+    return deletePhasesCommon.unBindNotRequired(ownServiceContext, SERVICE_NAME);
+}
+
+exports.unDeploy = function (ownServiceContext) {
+    // TODO
```
dsw88: Should this TODO be removed, or does it mean there's something left to do here?

lib/services/lambda/index.js
```diff
@@ -163,6 +163,9 @@ exports.consumeEvents = function (ownServiceContext, ownDeployContext, producerS
             principal = producerDeployContext.eventOutputs.principal;
             sourceArn = producerDeployContext.eventOutputs.eventRuleArn;
         }
+        else if (producerServiceType === 'alexaskillkit') {
+            principal = 'alexa-appkit.amazon.com';
```
dsw88: I think you'll need to provide a sourceArn here as well, but I'm not sure what that would be for the Alexa Skill Kit. It's been required on every other service I've done though.

ndpete: There is no sourceArn for the alexa skill kit. It just takes the prinicipal.

test/services/alexaskillkit/alexaskillkit-test.js
```diff
@@ -0,0 +1,66 @@
...
+            let serviceContext = {
+                params: {}
+            }
+            old_region = accountConfig.region
+            accountConfig.region = 'us-east-1' //toggle region
+            let errors = alexaSkillKit.check(serviceContext);
+            accountConfig.region = old_region //reset the region
+            expect(errors.length).to.equal(0);
+        });
+    });
```
dsw88: I know it's a bit more verbose, but it might be good to have some tests for the other phase contract methods, even though they don't really do much. I like knowing that they still don't do much farther down the road on each build, and if someone does change them then the breaking tests are a signal that other things might be affected too since they're changing contract methods.

lib/services/alexaskillkit/index.js
ndpete: Ah good catch. Thanks

dsw88:
This change looks great, I just had a few questions and comments. Looks pretty well good though!

lib/services/alexaskillkit/index.js
ndpete: Is it worth even logging that statement since I'm not actually producing any events? The permission gets created as part of the lambda's consume event phase.

dsw88: Yeah normally the "producer" would have something to do, but in this case it doesn't. I'd still like a message of some sort to indicate it isn't getting skipped, but maybe just change the message to something like this:
```
`${SERVICE_NAME} - Nothing to do while producing events from '${ownServiceContext.serviceName}' to consumer '${consumerServiceContext.serviceName}'
```

lib/services/lambda/index.js
dsw88: Cool sounds good!

test/services/alexaskillkit/alexaskillkit-test.js
ndpete: Thats a good point. I'm adding those now.

lib/services/alexaskillkit/index.js
ndpete: This has been resolved and refactored in the latest commit.

ndpete:
I've incorporated your comments in the latest commit.

dsw88:
Looks great!

PR #222, opened 2017-08-03
added support for dead-letter queues in the sqs service

saholman:


docs/supported-services/sqs.rst
```diff
@@ -75,13 +80,61 @@ This Handel file shows an SQS service being configured:
...
+            content_based_deduplication: true
+            delay_seconds: 2
+            max_message_size: 262140
+            message_retention_period: 345601
+            receive_message_wait_time_seconds: 4
+            visibility_timeout: 40
+
+.. _dead-letter:
+
+DeadLetterQueue
```
dsw88: Let's move this under the "Parameters" section since it's part of the parameters information. You can change the heading type to be under Parameters by using the next heading down, which would be the "~" character instead of the "-" character.

dsw88:
Other than my comment about the change to the docs, this PR looks good.

PR #226, opened 2017-08-07
Add custom descriptions

byujol:
Some minor changes to allow for custom descriptions for various components. The mocha tests all passed.


docs/supported-services/beanstalk.rst
```diff
@@ -37,6 +37,16 @@ Parameters
      - Yes
      - 
      - The ElasticBeanstalk solution stack you wish to use. This determines what AMI your application runs on. See `Elastic Beanstalk Supported Platforms <http://docs.aws.amazon.com/elasticbeanstalk/latest/dg/concepts.platforms.html>`_ for the list of solution stacks.
+   * - description_template
```
dsw88: I almost think that having two parameters in Beanstalk for a description might not provide as much value as mental overhead. What would you think about having a single "description" parameter that applies to both the template and the version? That way if in the future other parts of Beanstalk have descriptions we can just apply the single description to everything, and the configuration will be more consistent with the other services.

byujol:
Sounds good. I've merged the two descriptions for beanstalk environments.

PR #228, opened 2017-08-07
Add support for adding BinaryMediaTypes to apigateway

### LINKED ISSUE #223: API Gateway - Add ability to configure Binary Media
Currently you can't specify Binary Media in Handel, @bradleyreneer needs to do this for an API he's working on and currently has to configure it manually.

byujol:
This will add the BinaryMediaTypes support as requested.

Resolves #223

### DIFF
--- a/lib/services/apigateway/index.js
+++ b/lib/services/apigateway/index.js
@@ -84,7 +84,14 @@ function getCompiledApiGatewayTemplate(stackName, ownServiceContext, dependencie
     }
 
 		// Add binary media types if specified
-		if(serviceParams.binary_media_types) handlebarsParams.binaryMediaTypes=serviceParams.binary_media_types;
+		if(serviceParams.binary_media_types)
+    {
+      handlebarsParams.binaryMediaTypes=[];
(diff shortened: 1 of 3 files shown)


lib/services/apigateway/index.js
```diff
@@ -83,6 +83,16 @@ function getCompiledApiGatewayTemplate(stackName, ownServiceContext, dependencie
         tags: deployPhaseCommon.getTags(ownServiceContext)
     }
 
+		// Add binary media types if specified
+		if(serviceParams.binary_media_types)
+    {
+      handlebarsParams.binaryMediaTypes=[];
+      for(let type of serviceParams.binary_media_types)
```
dsw88: This won't work when the user hasn't specified binary_media_types, because serviceParams.binary_media_types will be undefined, which isn't iterable, so we should probably put this in an if statement checking if the user has defined binary_media_types.

dsw88: Wait I think I totally missed line 87, sorry about that, that's what I was looking for!

PR #236, opened 2017-08-14
130

byujol:
Removes s3 handel artifacts on delete. Also, renamed unDeployCloudFormationStack to unDeployService and fixed the action message in the bin/handel file.


lib/common/delete-phases-common.js
```diff
@@ -59,9 +60,16 @@ exports.unDeployCloudFormationStack = function (serviceContext, serviceType) {
             }
         })
         .then(() => {
+          return s3Calls.deleteMatchingPrefix('handel-'+accountConfig.region+'-'+accountConfig.account_id,serviceContext.appName+'/'+serviceContext.environmentName+'/'+serviceContext.serviceName);
+          return;
```
dsw88: This extra return should be removed, it will be treated as an error by linters since it's unreachable code.

lib/common/delete-phases-common.js
```diff
@@ -59,9 +60,16 @@ exports.unDeployCloudFormationStack = function (serviceContext, serviceType) {
             }
         })
         .then(() => {
+          return s3Calls.deleteMatchingPrefix('handel-'+accountConfig.region+'-'+accountConfig.account_id,serviceContext.appName+'/'+serviceContext.environmentName+'/'+serviceContext.serviceName);
```
dsw88: Could we split this out into two variables that get referenced in this function? Also can we use the ES6 string interpolation syntax? That will help it be a bit more understandable for people reading it after the fact I think. Here's what I'm thinking of:

```
let bucketName = `handel-${accountConfig.region}-${accountConfig.account_id}`;
let prefix = `${serviceContext.appName}/${serviceContext.environmentName}/${serviceContext.serviceName}`;
return s3Calls.deleteMatchingPrefix(bucketName, prefix);      
```

test/aws/s3-calls-test.js
```diff
@@ -21,6 +21,8 @@ const s3Calls = require('../../lib/aws/s3-calls');
 const sinon = require('sinon');
 const childProcess = require('child_process');
 
+console.log('args',process.argv);
```
dsw88: I assume this is something left over from debugging that should be removed.

lib/common/delete-phases-common.js
```diff
@@ -59,9 +60,16 @@ exports.unDeployCloudFormationStack = function (serviceContext, serviceType) {
...
+          return;
+        })
+        .then(() => {
             return new UnDeployContext(serviceContext);
         });
 }
+exports.unDeployCloudFormationStack = function(serviceContext,serviceType) {
```
dsw88: I would vote for removing this little method since you've already switched over the usages of "unDeployCloudFormationStack" to "unDeployService".

PR #238, opened 2017-08-15
Validate aws authentication prior to deploy or delete

byujol:
Also added a fix to the delete phase s3 cleanup.


lib/aws/iam-calls.js
```diff
@@ -258,4 +258,25 @@ exports.constructPolicyDoc = function (policyStatements) {
...
-}
\ No newline at end of file
+}
+
+/**
+ * List roles
+ * 
+ * This method is used to determine the account id
+ */
+exports.showAccount = function()
```
dsw88: It looks like we're missing a unit test on this function.

lib/common/util.js
```diff
@@ -242,4 +243,16 @@ exports.deleteFolderRecursive = function(path) {
     });
     fs.rmdirSync(path);
   }
-};
\ No newline at end of file
+};
+
+exports.validateCredentials = function(accountConfig) {
```
dsw88: This also looks like it needs a unit test.

PR #243, opened 2017-08-16
beanstalk Dockerrun.aws.json tags

byujol:
Added <aws_account_id> and <aws_region> tag replacment for beanstalk Dockerrun.aws.json files

### DIFF
--- a/lib/common/util.js
+++ b/lib/common/util.js
@@ -24,6 +24,16 @@ const IAM = require('../aws/iam-calls');
 const ncp = require('ncp').ncp;
 ncp.limit = 16;
 
+exports.readDirSync = function (filePath) {
+    try {
+        return fs.readdirSync(filePath);
+    }
+    catch (e) {
+        winston.error("Couldn't read directory: " + e);
+        return null;
+    }
+};
+
 exports.readFileSync = function (filePath) {
     try {
         return fs.readFileSync(filePath, 'utf8');
@@ -32,7 +42,18 @@ exports.readFileSync = function (filePath) {
         winston.error("Couldn't load file: " + e);
         return null;
     }
-}
+};
+
+exports.writeFileSync = function (filePath, data) {
+    try {
+        fs.writeFileSync(filePath, data);
+        return data;
+    }
+    catch (e) {
+        winston.error("Couldn't write file: " + e);
+        return null;
+    }
+};
 
 /**
  * Reads the contents of a YAML file in a synchronous manner. Don't
@@ -89,29 +110,10 @@ exports.readYamlFileAsync = function (filePath) {
  * @returns {Promise.<String>} - A Promise of a String representing the contents of fileName with substitutions
  */
 exports.replaceTagInFile = function (listTag, filePath, fileName) {
-  return new Promise((resolv,reject)=>
-  {
-    fs.readFile(`${filePath}/${fileName}`,'utf8',(err,readData)=>
-    {
-      if(err) reject(err);
-      else    resolv(readData);
-    });
-  })
-  .then(parm=>
-  {
-    for(let tag of listTag) parm=parm.replace(tag.regex,tag.value);
-    return new Promise((resolv,reject)=>
-    {
-      fs.writeFile(`${filePath}/${fileName}`,parm,(err,rc)=>
-      {
-        if(err) reject(err);
-        else
-        {
-          resolv(parm);
-        }
-      });
-    });
-  });
+  let readData=exports.readFileSync(`${filePath}/${fileName}`);
+  if(!readData) return readData;
+  for(let tag of listTag) readData=readData.replace(tag.regex,tag.value);
+  return exports.writeFileSync(`${filePath}/${fileName}`,readData);
 };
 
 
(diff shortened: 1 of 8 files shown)


lib/common/util.js
```diff
@@ -80,6 +80,43 @@ exports.readYamlFileAsync = function (filePath) {
...
+ * Takes the given directory path and file and replace tags found in the file with
+ *   values from the tag list
+ * 
+ * @param {Array} listTag - An array of objects containing fields regex: {regex} and value: {String}
+ * @param {String} filePath - The full path to the file on disk to do substitutions to
+ * @param {String} fileName - The name of the file on disk
+ * @returns {Promise.<String>} - A Promise of a String representing the contents of fileName with substitutions
+ */
+exports.replaceTagInFile = function (listTag, filePath, fileName) {
+  return new Promise((resolv,reject)=>
```
dsw88: Could we change this to use readFileSync and writeFileSync? I know those are blocking calls, but Handel doesn't need to worry too much about that since we don't have multiple users being serviced in a web service, and the predominant time spent in the library is waiting for AWS resources to create. I like readFileSync and writeFile sync because we can reduce this function to a much more readable few lines rather than using callbacks and promises.

lib/services/beanstalk/deployable-artifact.js
```diff
@@ -33,62 +35,120 @@ function uploadDeployableArtifactToS3(serviceContext, fileToUpload, fileExtensio
...
-                        });
-                })
-                .then(s3ArtifactInfo => {
-                    util.deleteFolderRecursive(tempDirPath); //Delete the whole temp folder
-                    return s3ArtifactInfo;
-                });
+    .then(()=>
+    {
+      // If one of the files in the directory is a Dockerrun.aws.json file then do marker substitution
+      return new Promise((resolve,reject)=>
```
dsw88: Likewise, can we refactor a bit to use synchronous file calls here? That will make this a lot more readable too, and we have the luxury of doing that since this is a CLI tool that spends most of its time waiting for AWS.

lib/services/beanstalk/deployable-artifact.js
```diff
@@ -33,62 +35,120 @@ function uploadDeployableArtifactToS3(serviceContext, fileToUpload, fileExtensio
...
+    if (fs.existsSync(userEbextensionsDir)) {
+      return util.copyDirectory(userEbextensionsDir, `${tempDirPath}/.ebextensions/`);
+    }
+    else {
+      return true;
+    }
+  })
+  .then(()=>
+  {
+    if(fileName == "Dockerrun.aws.json")
```
dsw88: Likewise, can we refactor a bit to use synchronous file calls here in these file changes too? That will make this a lot more readable too, and we have the luxury of doing that since this is a CLI tool that spends most of its time waiting for AWS.

lib/services/beanstalk/deployable-artifact.js
```diff
@@ -101,10 +161,10 @@ function prepareAndUploadWar(pathToArtifact, ebextensionsToInject) {
     return Promise.reject(new Error("Not Implemented"));
 }
 
-function prepareAndUploadJar(ownServiceContext, pathToArtifact, ebextensionsToInject) {
+function prepareAndUploadJar(accountConfig, ownServiceContext, pathToArtifact, ebextensionsToInject) {
```
dsw88: Rather than passing accountConfig around, you can just require it at the top of your file like this:

```
const accountConfig = require('../../common/account-config')().getAccountConfig();
```

Then you can use it in any function in the file that needs it.

lib/services/beanstalk/index.js
```diff
@@ -295,7 +295,7 @@ exports.deploy = function (ownServiceContext, ownPreDeployContext, dependenciesD
         .then(serviceRole => {
             return getSystemInjectedEbExtensions(stackName, ownServiceContext, dependenciesDeployContexts)
                 .then(ebextensionFiles => {
-                    return deployableArtifact.prepareAndUploadDeployableArtifact(ownServiceContext, ebextensionFiles)
+                    return deployableArtifact.prepareAndUploadDeployableArtifact(accountConfig, ownServiceContext, ebextensionFiles)
```
dsw88: Likewise in this file, you don't need to pass accountConfig into the method, just require it inside that file.

test/services/beanstalk/deployable-artifact-test.js
```diff
@@ -56,7 +56,7 @@ describe('deployable artifact module', function () {
             let deleteEbextensionsStub = sandbox.stub(ebextensions, 'deleteAddedEbExtensionsFromDirectory').returns(true);
             let copyDirectoryStub = sandbox.stub(util, 'copyDirectory').returns(Promise.resolve(true));
 
-            return deployableArtifact.prepareAndUploadDeployableArtifact(serviceContext, [])
+            return deployableArtifact.prepareAndUploadDeployableArtifact(accountConfig,serviceContext, [])
```
dsw88: Once you remove the accountConfig parameters in these methods you can remove their references here.

PR #244, opened 2017-08-16
Route53 Support

### LINKED ISSUE #237: Route53 Support
Here's what I'm proposing:

One - a new resource type, `route53zone`, representing a hosted zone in Route 53.  It would look something like this:

```yaml
my-hosted-zone:
  type: route53zone
  name: mydomain.byu.edu
  private: true #default to false
  vpcs: #required if private=true, cannot be specified if private=false.
    - vpc12312312
    - <account> #Uses default vpc from account config
```

Then, we can enhance the configuration of types that are valid DNS alias targets with a DNS configuration.  Using an ECS resource as an example, it could look like:

```yaml
my-ecs:
  type: ecs
  ...
  dns:
    name: mysubdomain.byu.edu
    protocols: [ ipv4, ipv6 ] #Or should we change this to be record_types: [ A, AAAA ]?
    comment: some comment
  dependencies:
    - my-hosted-zone
```

This would create A and AAAA records on the specified hosted zone, depending on which protocols are enabled. The default should be both, but may vary depending on what's supported by the service they're fronting. I could even make a case for us not giving the user any choice, and just setting up whichever records are supported by the backend service. 
The records would, in this case, point to the load balancer that is created to front the ECS service.

Here's a list of current service that could receive this enhancement:

* API Gateway (this would require adding support for custom domains)
* Beanstalk
* ECS
* S3 Static Site (this might require adding cloudfront support)

Theoretically, we could also expose RDS instances too, but I think we should always limit those to private zones.

As an additional rule, we could say that public zones (the default) must be a subdomain of byu.edu, but private zones can be whatever (since they just set up DNS rules in the specified VPCs).

If this looks good to you, I can start working on implementing it.  I'd like to get it out there soon, so that I can take advantage of it.

ThatJoeMoore:
Adds support for a `route53zone` service and for configuring DNS records for Beanstalk and ECS.

Resolves #237 .

The only outstanding thing (which should probably be a separate PR) is adding the ability for services to give messages to the user (separate from the normal logging), so that things like the manual steps for `route53zone` can be explained.

### DIFF
--- a/docs/supported-services/beanstalk.rst
+++ b/docs/supported-services/beanstalk.rst
@@ -72,11 +72,6 @@ Parameters
      - No
      - 
      - Any user-specified environment variables to inject in the application.
-   * - dns
-     - List<:ref:`route53zone-records`>
-     - No
-     -
-     - DNS records to create and associate with the load balancer for this application.
    * - tags
      - :ref:`beanstalk-tags`
      - No
@@ -140,6 +135,12 @@ The Routing element is defined by the following schema:
     routing:
       type: <http|https>
       https_certificate # Required if you select https as the routing type
+      dns_names:
+       - <string> # Optional
+
+DNS Names
+`````````
+The `dns_names` section creates one or more dns names that point to this load balancer. See :ref:`route53zone-records` for more.
 
 .. _beanstalk-tags:
 
--- a/docs/supported-services/beanstalk.rst
+++ b/docs/supported-services/beanstalk.rst
@@ -138,8 +138,6 @@ The Routing element is defined by the following schema:
       dns_names:
        - <string> # Optional
 
-DNS Names
-`````````
 The `dns_names` section creates one or more dns names that point to this load balancer. See :ref:`route53zone-records` for more.
 
 .. _beanstalk-tags:
--- a/docs/supported-services/beanstalk.rst
+++ b/docs/supported-services/beanstalk.rst
@@ -72,6 +72,11 @@ Parameters
      - No
      - 
      - Any user-specified environment variables to inject in the application.
+   * - dns
+     - List<:ref:`route53zone-records`>
+     - No
+     -
+     - DNS records to create and associate with the load balancer for this application.
    * - tags
      - :ref:`beanstalk-tags`
      - No
--- a/docs/supported-services/ecs.rst
+++ b/docs/supported-services/ecs.rst
@@ -182,6 +182,12 @@ The `load_balancer` section is defined by the following schema:
       type: <string> # Required. Allowed values: `http`, `https`. 
       timeout: <integer> # Optional. The connection timeout on the load balancer
       https_certificate: <string> # Required if type=https. The ID of the ACM certificate to use on the load balancer.
+      dns_names:
+       - <string> # Optional.
+
+DNS Names
+`````````
+The `dns_names` section creates one or more dns names that point to this load balancer. See :ref:`route53zone-records` for more.
 
 .. _ecs-tags:
 
--- a/docs/supported-services/ecs.rst
+++ b/docs/supported-services/ecs.rst
@@ -53,6 +53,11 @@ Parameters
      - No
      - 
      - If your task needs routing from a load balancer, this section can be used to configure the load balancer's options.
+   * - dns
+     - List<:ref:`route53zone-records`>
+     - No
+     -
+     - DNS records to create and associate with the load balancer for this task.
    * - tags
      - :ref:`ecs-tags`
      - No
--- a/docs/supported-services/ecs.rst
+++ b/docs/supported-services/ecs.rst
@@ -185,8 +185,6 @@ The `load_balancer` section is defined by the following schema:
       dns_names:
        - <string> # Optional.
 
-DNS Names
-`````````
 The `dns_names` section creates one or more dns names that point to this load balancer. See :ref:`route53zone-records` for more.
 
 .. _ecs-tags:
--- a/docs/supported-services/route53zone.rst
+++ b/docs/supported-services/route53zone.rst
@@ -1,0 +1,209 @@
+.. _route53zone:
+
+Route 53 Hosted Zone
+====================
+This document contains information about the Route 53 Hosted Zone service supported in Handel. This Handel service provisions a Route 53 Hosted Zone, in which you can create other DNS records.
+
+Service Limitations
+-------------------
+The following Route 53 features are not currently supported in this service:
+
+* Domain Name Registration - all public zones must be subdomains of byu.edu.
+
+Parameters
+----------
+.. list-table::
+   :header-rows: 1
+
+   * - Parameter
+     - Type
+     - Required
+     - Default
+     - Description
+   * - type
+     - string
+     - Yes
+     -
+     - This must always be *route53zone* for this service type.
+   * - name
+     - string
+     - Yes
+     -
+     - The DNS name for this hosted zone. If this is a public zone, it must end with .byu.edu.
+   * - private
+     - boolean
+     - No
+     - false
+     - Whether or not this is a private zone. If it is a private zone, it is only accessible by the VPCs listed below
+   * - vpcs
+     - List<string>
+     - No
+     - [ <account> ]
+     - List of VPC ids. Must only be specified if this is a private zone. The special value <account> references the VPC listed in the account configuration.
+   * - tags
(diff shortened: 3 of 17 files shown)


ThatJoeMoore:
Please note that I haven't tested this with a real app yet, and I'm leaving on vacation in a few minutes, so.... ¯\_(ツ)_/¯

docs/supported-services/beanstalk.rst
```diff
@@ -134,6 +134,12 @@ The Routing element is defined by the following schema:
     routing:
       type: <http|https>
       https_certificate # Required if you select https as the routing type
+      dns_names:
+       - <string> # Optional
+
+DNS Names
```
dsw88: The DNS Names header looks just about the same size as its enclosing Routing header when rendered. What would you say to removing the DNS Names header? To me it looks more clearly like it belongs in the Routing section and not on its own.

docs/supported-services/ecs.rst
```diff
@@ -176,6 +181,12 @@ The `load_balancer` section is defined by the following schema:
       type: <string> # Required. Allowed values: `http`, `https`. 
       timeout: <integer> # Optional. The connection timeout on the load balancer
       https_certificate: <string> # Required if type=https. The ID of the ACM certificate to use on the load balancer.
+      dns_names:
+       - <string> # Optional.
+
+DNS Names
```
dsw88: When I tried to build the docs I got an error about this DNS Names header, and it didn't get rendered. Like in Beanstalk's docs, I actually prefer it without the DNS Names header, so my vote would be remove it if you're ok with that.

docs/supported-services/beanstalk.rst
ThatJoeMoore: Yeah, that makes sense. I think that header is an artifact from when the DNS config was more complex.

I can make that change later today.

docs/supported-services/route53zone.rst
```diff
@@ -0,0 +1,174 @@
...
+-------------------
+The following Route 53 features are not currently supported in this service:
+
+* Domain Name Registration
+
+Manual Steps
+------------
+If creating a public zone as a subdomain of another domain (like myapp.byu.edu), you must register it with your DNS provider.
+
+For BYU users, please refer to `this document <https://byuoit.atlassian.net/wiki/spaces/OAPP/pages/40075276/Routing+BYU+DNS+into+AWS>`_.
```
dsw88: We're trying to keep anything BYU specific out of Handel (other than the fact that we wrote it), so I'd like to remove this sentence. How about this section to be something more generic like the following:

```
If you are creating a public zone as a subdomain of another domain (like myapp.mydomain.com), you must register it with your DNS provider. 

If you are using Handel for your work at a company or organization of some kind, they likely have a process for registering these hosted zones with their DNS provider. Check with the networking groups in your organization to find out how you can do this.
```

That's just an example that could probably be improved of course, but we would like to keep anything BYU-special out of the library.

docs/supported-services/route53zone.rst
```diff
@@ -0,0 +1,174 @@
...
+    environments:
+      dev:
+        public-zone:
+          type: route53zone
+          name: mydomain.byu.edu
+          tags:
+            mytag: mytagvalue
+        private-zone:
+          type: route53zone
+          name: private.myapp # Doesn't have to be a .byu.edu domain
```
dsw88: Likewise, maybe let's make the example domains be a bit more generic like myapp.myfakedomain.com. Although now that I'm harping on this, I see I haven't followed that example on another page. :) I'll go remove the BYU-specific stuff there too.

docs/supported-services/route53zone.rst
```diff
@@ -0,0 +1,174 @@
...
+          private: true
+          tags:
+            mytag: mytagvalue
+
+Depending on this service
```
dsw88: I hadn't thought of making this service a dependency for others, but I guess there could definitely be use cases for people injecting these variables in their apps.

docs/supported-services/route53zone.rst
```diff
@@ -0,0 +1,174 @@
...
+
+Certain supported services can create an alias record in this zone.  The currently supported services are:
+
+* Beanstalk
+* ECS
+
+Each service can support multiple DNS entries. See the individual service documentation for how to define the DNS names.
+
+The DNS name must either match or be a subdomain of an existing Route 53 hosted zone name. If the hosted zone is configured
+in the same Handel environment, you must declare it as a dependency of the service consuming it.
```
dsw88: Just curious, what is the motivation for requiring that the hosted zone be declared as a dependency of the service using that hosted zone for a DNS name? If the hosted zone was configured some other way outside of the Handel file, we still allow them to do it right?

lib/services/beanstalk/index.js
```diff
@@ -272,10 +305,18 @@ function getSystemInjectedEbExtensions(stackName, ownServiceContext, dependencie
 
 exports.check = function (serviceContext) {
     let errors = [];
+    let params = serviceContext.params;
 
     //TODO - Implement check method
```
dsw88: Ah I see here I've forgotten to implement check for the Beanstalk method. :) Thanks for adding the check for the DNS names stuff! I'll try to improve this with the remaining checks at some point.

lib/services/ecs/index.js
```diff
@@ -45,8 +46,8 @@ function getLatestEcsAmiId() {
 
 
 function getCompiledEcsTemplate(stackName, clusterName, ownServiceContext, ownPreDeployContext, dependenciesDeployContexts, userDataScript, ecsServiceRole) {
-    return getLatestEcsAmiId()
-        .then(latestEcsAmi => {
+    return Promise.all([getLatestEcsAmiId(), route53.listHostedZones()])
+        .then(([latestEcsAmi, hostedZones]) => {
```
dsw88: Just curious, is this an ES6 syntax that I haven't used before? I notice you're just referencing "hostedZones" by itself below, even though it appears to be the second element in an array.

lib/services/route53zone/index.js
```diff
@@ -0,0 +1,141 @@
...
+
+exports.unDeploy = function (ownServiceContext) {
+    return deletePhasesCommon.unDeployService(ownServiceContext, SERVICE_NAME);
+}
+
+exports.producedEventsSupportedServices = []; //TODO - No events supported yet, but we will support some like Lambda
+
+exports.producedDeployOutputTypes = [
+    'environmentVariables',
+    'policies'
```
dsw88: This one doesn't output any policies for consumers, correct? If so, let's remove 'policies' from this list, because this list is used when determining which services can be dependencies of each other.

dsw88:
This PR looks great! I just had a few comments and questions, and I want to test this out on a real example before merging.

dsw88:
On a side note, I don't know why I haven't thought of this before, but as soon as we merge this PR, I'm going to remove the requirement that every service deployer implement every contract method. I can just have the framework check to see if the service deployer has that method, and if so call it. If it doesn't implement a particular method, the framework will just provide an empty context in the same way service deployers are doing today.

That will really remove a lot of redundant code from the service deployers and their tests for the phases they don't implement.

docs/supported-services/route53zone.rst
ThatJoeMoore: Yeah, the use case I thought of is if you need to know what your public-facing URL is (like to do CAS redirects), you can avoid hardcoding it by just depending on your hosted zone.

docs/supported-services/route53zone.rst
ThatJoeMoore: It's all about the ordering.  If the hosted zone is already set up, we have no worries.  However, if it's in the same Handel file, we need to make sure that we set up the hosted zone before the ECS or Beanstalk app.

lib/services/ecs/index.js
ThatJoeMoore: Yeah, ES6/Node 6 added destructuring for arrays and objects:

`let [hi, hello] = ['ciao', 'buongiorno'];`
`let {foo, bar} = {foo: 1, bar: 2};`

I think I'm going to change this up a bit to make it more readable to someone who doesn't spot the destructuring.

lib/services/route53zone/index.js
ThatJoeMoore: Yay copypasta!

ThatJoeMoore:
Alright, I think I've addressed everything.

PR #249, opened 2017-08-21
handel now supports dynamodb streams

saholman:


docs/supported-services/dynamodb.rst
```diff
@@ -192,9 +196,25 @@ The DynamoDB service outputs the following environment variables:
...
+    batch_size: <number> # Optional.  Default: 100
+
+BatchSize
+~~~~~~~~~~~~
+The largest number of records that AWS Lambda will retrieve from your event source at the time of invoking your function. Your function receives an event with all the retrieved records. The default is 100 records.
+
 Events produced by this service
 -------------------------------
-The DynamoDB service does not currently produce events for other Handel services to consume. Support for events to services such as Lambda is planned to be added in the future.
+The SNS service currently produces events for the following services types:
```
dsw88: Small copy/paste error here. It should be "DynamoDB" instead of "SNS"

saholman: Fixed with commit 0cd70c4

lib/services/lambda/index.js
```diff
@@ -99,6 +101,9 @@ function getPolicyStatementsForLambdaRole(serviceContext, dependenciesDeployCont
     return deployPhaseCommon.getAllPolicyStatementsForServiceRole(ownPolicyStatements, dependenciesDeployContexts);
 }
 
+function constructRoleName(ownServiceContext) {
```
dsw88: I think you can get rid of this function and use deployPhaseCommon.getResourceName instead. That should be the same as this function, and everyone else uses it, so if we ever change those names it'll be nice to only change that in one place.

lib/services/lambda/index.js
```diff
@@ -150,37 +155,70 @@ exports.deploy = function (ownServiceContext, ownPreDeployContext, dependenciesD
...
-
-        return lambdaCalls.addLambdaPermissionIfNotExists(functionName, principal, sourceArn)
-            .then(permissionStatement => {
+            let tableStreamArn = producerDeployContext.eventOutputs.tableStreamArn;
+            let tableName = producerDeployContext.eventOutputs.tableName;
+            let lambdaConsumers = producerDeployContext.eventOutputs.lambdaConsumers;
+            let lambdaConsumer;
+            let policyStatementsToConsume = JSON.parse(util.readFileSync(`${__dirname}/lambda-dynamodb-stream-role-statements.json`));
+            policyStatementsToConsume[0].Resource = [];
+            policyStatementsToConsume[0].Resource.push(`arn:aws:dynamodb:${accountConfig.region}:${accountConfig.account_id}:table/${tableName}/stream/*`)
```
dsw88: Can you use tableStreamArn here instead of hard-coding this string? They look like they should be the same thing.

lib/services/lambda/index.js
```diff
@@ -150,37 +155,70 @@ exports.deploy = function (ownServiceContext, ownPreDeployContext, dependenciesD
...
+            return iamCalls.attachStreamPolicy(constructRoleName(ownServiceContext), policyStatementsToConsume)
+            .then(() => {
                 winston.info(`${SERVICE_NAME} - Allowed consuming events from ${producerServiceContext.serviceName} for ${ownServiceContext.serviceName}`);
-                return resolve(new ConsumeEventsContext(ownServiceContext, producerServiceContext));
-            });
+                return null;
```
dsw88: You should be able to just leave this "return null;" off if you don't want to return anything and it will return undefined.

lib/services/lambda/index.js
saholman: The problem is that this has the * after it, where as the tableStreamArn has a datestamp after /stream/

dsw88: Ah yeah I see that this is returning back from the DynamoDB CF stack. How about improving it a bit to use the tableStreamArn variable, but strip off the datestamp at the end and put a "*" on it instead? That way you know that the rest of the ARN is correct and don't have to manually construct it entirely. I think it helps reduce the risk of someone coming along later and accidentally changing it incorrectly.

lib/services/lambda/index.js
saholman: resolved by commit ccb31d7

lib/services/lambda/index.js
saholman: resolved by commit ccb31d7

lib/services/lambda/index.js
saholman: resolved by commit ccb31d7

PR #250, opened 2017-08-21
add S3 lifecycles support

### LINKED ISSUE #169: S3 - Support bucket lifecycle configuration
Currently the S3 Handel service doesn't let you configure any bucket lifecycles. We should allow this feature (including potentially the built-in integration with Glacier).

ndpete:
This add initial support for s3 bucket lifecycle policies

resolves #169

### DIFF
--- a/docs/supported-services/s3.rst
+++ b/docs/supported-services/s3.rst
@@ -13,7 +13,6 @@ Service Limitations
 This service currently only provisions a bare-bones S3 bucket for data storage. It does support versioning, but the following other features are not currently supported:
 
 * CORS configuration
-* Bucket lifecycle
 * Bucket logging
 * Cross-region replication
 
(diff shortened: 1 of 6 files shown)


docs/supported-services/s3.rst
```diff
@@ -54,12 +53,80 @@ This service takes the following parameters:
...
+     - Lifecycle Policies to apply to the bucket. See `AWS Docs for more info <http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-lifecycleconfig-rule.html>`_
    * - tags
      - :ref:`s3-tags`
      - No
      -
      - Any tags you want to apply to your S3 bucket
 
+.. _s3-lifecycles:
+
+lifecycles
```
dsw88: I usually try to capitalize these section headers, so "Lifecycles"

docs/supported-services/s3.rst
```diff
@@ -54,12 +53,80 @@ This service takes the following parameters:
...
+~~~~~~~~~~
+A list of life cycle rules
```
dsw88: It looks like you're missing docs for the "prefix" field. Is that on the "lifecycles" section or under each "transitions" section?

ndpete: Oh good catch, thats on the Lifecycles section right under name before transitions most likely.

ndpete: I'll clean those up if you'd like

PR #257, opened 2017-08-24
Now supports lambdas inside vpc

saholman:


docs/supported-services/lambda.rst
```diff
@@ -56,6 +55,11 @@ Parameters
      - No
      - 3
      - The timeout in seconds for your function. Max 300
+   * - vpc
+     - boolean
+     - No
+     - false
+     - If true, the lambda will be deployed inside the VPC and private subnets defined in the account config file.
```
dsw88: I propose changing this sentence to: "If true, the lambda will be deployed inside your VPC. Inside your VPC, it will be able to communicate with resources like RDS databases and ElastiCache clusters".

I don't think they need to worry too much about the private subnets and account config file in the service docs, it's mostly just important for them to know what effect that will have on what they can do once it's inside their VPC.

lib/phases/check.js
```diff
@@ -23,7 +23,13 @@ exports.checkServices = function (serviceDeployers, environmentContext) {
     let errors = [];
     _.forEach(environmentContext.serviceContexts, function (serviceContext) {
         let serviceDeployer = serviceDeployers[serviceContext.serviceType];
-        let checkErrors = serviceDeployer.check(serviceContext);
+        let dependenciesServiceContexts = [];
```
dsw88: Maybe split this out into a little function in this file called "getDependenciesServiceContexts". That will help reduce the noise in this function.

saholman:
I didn't test redeploying something, but it should be exactly the same thing.

PR #259, opened 2017-08-24
Added vpc support for ApiGateway

saholman:


dsw88:
Let's actually wait to merge Lambda first, because I see you have duplicate changes in this one, so we'll have multiple merge conflicts after Lambda is merged.

PR #264, opened 2017-08-25
Add support for KMS

ThatJoeMoore:
This adds support for creating and depending on KMS keys.

By default, it creates an alias for the key, named `alias/<appName>/<envName>/<serviceName>`.

**Future Enhancements:**

- Add support for publishing events whenever the key is rotated.  I'll probably tackle this as part of my work on Payment Manager v2.


dsw88:
Looks great, merged. Thanks so much for adding this!

PR #268, opened 2017-08-30
Add ECS CloudWatch logging

### LINKED ISSUE #267: Add CloudWatch logging to ECS containers
I was just going to implement this as a pull request, but I've come across a design issue centering around the creation of log groups for the containers.  Specifically, how to configure the log retention time

# Option 1: Implicit log group per ECS cluster

```yml
    type: ecs
    containers:
    - name: app
      ...
      logging:
        enabled: true
        prefix: app # defaults to container name
    log_retention_days: 14 # Default to infinite?
```

Pro: single log group
Con: It's weird to have logging in two places.  

# Option 2: Define groups in container

```yml
    type: ecs
    containers:
    - name: app
      ...
      logging:
        enabled: true
        group: mygroup #defaults to cluster name
        prefix: app # defaults to container name
```

Pro: More configurable.  Each container can log to its own group.
Con: How do we configure the retention time?  We could do it at the container level, but if more than one container wants to share the log group, which configuration do we prefer?  Or do we force them to create unique log groups?

# Option 3: Make logging on/off, no customization

```yml
    type: ecs
    ...
    logging: enabled
    log_retention_days: 30
```

There would be one log group, with no control over the name (which would be computed). Each container would get a log prefix matching its name.

Pro: Simple
Con: Not very flexible.  But do we need much flexibility?

# Option 4: Add logging service type

```yml
log-group:
  type: loggroup
  name: my-log-group
  retention_in_days: 30
app:
  type: ecs
  containers:
  - name: app
    ...
    logging:
      enabled: true
      group: my-log-group
      prefix: app
  dependencies:
  - log-group #If we don't create the log group before the cluster, it'll spin and spin and never complete.  Don't ask me how I know.
```
Pros: Lots of control, can be used by other services
Cons: Complicated configuration

What do you think?  The interesting part here is that logging is configured at the container level, but the log groups need to exist before the container can successfully start.

ThatJoeMoore:
Fixes #267 . 

Logging is turned on by default.

### DIFF
--- a/docs/supported-services/ecs.rst
+++ b/docs/supported-services/ecs.rst
@@ -53,6 +53,16 @@ Parameters
      - No
      - 
      - If your task needs routing from a load balancer, this section can be used to configure the load balancer's options.
+   * - logging
+     - string
+     - No
+     - "enabled"
+     - Turns CloudWatch logging on or off. Must be either "enabled" or "disabled". See :ref:`ecs-logging` for more.
+   * - log_retention_in_days
+     - number
+     - No
+     -
+     - Configures the log retention duration for CloudWatch logs. If not set, logs are kept indefinitely.
    * - tags
      - :ref:`ecs-tags`
      - No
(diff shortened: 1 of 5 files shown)


docs/supported-services/ecs.rst
```diff
@@ -53,6 +53,16 @@ Parameters
...
      - If your task needs routing from a load balancer, this section can be used to configure the load balancer's options.
+   * - logging
+     - string
+     - No
+     - "enabled"
+     - Turns CloudWatch logging on or off. Must be either "enabled" or "disabled". See :ref:`ecs-logging` for more.
+   * - log_retention_in_days
+     - number
+     - No
+     -
```
dsw88: Can you put the default value for log retention here in this column?

ThatJoeMoore: Yeah, that had just occurred to me.  I've now got it defaulting to '0', which means 'indefinite'.

dsw88:
Thanks for implementing this, it's been sorely needed in the ECS service!

ThatJoeMoore:
@dsw88 Yeah, I'm just kind of working through one of my apps and adding everything I need as I come to it :)

PR #271, opened 2017-09-02
Better env vars

### LINKED ISSUE #270: Modify naming for environment variables and parameters store parameters
For env vars remove the first 3 parts of the prefix.  They are not needed since they are only shown to one app at a time.

For parameter store parameters inject an env var called something like HANDEL_PS_PREFIX that contains the parameter store prefix so developers don't have to remember how to construct the prefix.

This simplification should remove the need to write SDKs like https://www.npmjs.com/package/handel-utils for each language that we use.

We need to be careful about backwards compatibility.

ThatJoeMoore:
Resolves #270 .

Adds `HANDEL_PARAMETER_STORE_PREFIX`, with the computed prefix for parameter store values.

Changes injected environment variables to drop the service type, app name, and env name.  Environment variables now look like `<SERVICE_NAME>_<VARIABLE_NAME>`.

For compatibility, the old variable names are still included.  They will be removed pre-1.0 release.

Question to consider: do we want to prefix the environment variables with something like `HANDEL_`? I personally don't think it's necessary, but I can see how we might want the extra namespacing.

### DIFF
--- a/docs/handel-basics/consuming-service-dependencies.rst
+++ b/docs/handel-basics/consuming-service-dependencies.rst
@@ -31,10 +31,10 @@ The following Handel file defines a Beanstalk service that depends on an SQS que
 
 Handel will inject environment variables in the Beanstalk application for the SQS queue, such as the queue's ARN, name, and URL. You can read these environment variables when you are writing code to communicate with the queue.
 
-.. _environment-variable-prefix:
+.. _environment-variable-names:
 
-Environment Variable Name
--------------------------
+Environment Variable Names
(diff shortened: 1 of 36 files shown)


dsw88:
I don't think the service env variables really need to be prefixed with HANDEL_, although I don't suppose it would hurt if they were. I'm fine mergine this as-is without the HANDEL_ prefix.

dsw88:
I'd like to remove the old style, so I may go see how many PRs I have to submit for people to use the new naming style. :)

PR #281, opened 2017-09-21
Ecs drain

byujol:
This change automatically adds a lifecycle hook to ecs instances deployed using handel to delays their termination for 300 seconds. During this lifecycle phase the HandelEcsDrainingLambda sets the status of the terminating ec2 instance to DRAINING causing future web requests to be serviced by other instances. There was also a modification to the HandelEcsAutoScalingLambda to only consider ec2 instances in an ACTIVE state during the evaluation of cluster capacity.


lib/services/ecs/ecs-service-template.yml
```diff
@@ -161,6 +161,13 @@ Resources:
       EvaluationPeriods: '1'
       Namespace: Handel/ECS
       Threshold: '0'
+  EcsAutoScalingLifecycle:
+    Type: AWS::AutoScaling::LifecycleHook
```
dsw88: I've not used these lifecycle hooks myself before. How does this invoke the draining Lambda you've defined?

byujol: It doesn't. It just causes the termination process to pause until it either receives a continue or times out. Currently I'm just letting it timeout. When the autoscaling group sets an ec2 instance to terminate the draining lambda gets invoked marking putting the ec2 instance into the draining status.

Issue #218, opened 2017-07-31
ECS - Automatically roll instances when the launch configuration is updated


dsw88:
Currently, if you update something like instance type, the instances in the auto-scaling group do not get automatically rolled. 

This change would roll those in a deploy when the launch configuration is updated.

dsw88:
Resolved by https://github.com/byu-oit/handel/pull/286

Issue #185, opened 2017-06-21
Adding key_name to existing ecs resource doesn't work


joshgubler:
Added a key_name to an existing ecs resource. Deployment log says that the CF stack was updated successfully, but the EC2 instance wasn't swapped out for a new instance with the key pair associated.

dsw88:
So I don't believe ECS has any built-in logic to rotate your instances when your launch configuration is changed. This means that when you update your key, you currently have to manually roll your instances by terminating them one-by-one. New ones will come up with the correct configuration.

ElasticBeanstalk does this for you, but ECS does not, so we might need to build something into Handel to support this automatically.

dsw88:
Resolved by https://github.com/byu-oit/handel/pull/286

PR #286, opened 2017-09-28
Ecs roll instances on update

byujol:
This change will cause all ec2 instances allocated to a cluster to roll on update. This will facilitate graceful cluster updates to changes in the ssh keys.


byujol:
This will fix issue #218 and #185

byujol:
Added code to only roll the entire cluster when either the instance type or ssh key has changed in the handel file.

PR #298, opened 2017-10-13
SES

brettbeatty:
So far it just asks SES to verify an email address if it has not been verified with the current AWS account and gives dependent services permission to send email from that email address.

To allow multiple applications to share an email address, I don't have it deleting email addresses when the service is deleted. I have that warning in the documentation.

### DIFF
--- a/lib/aws/ses-calls.js
+++ b/lib/aws/ses-calls.js
@@ -23,7 +23,7 @@ function verifyEmailAddress(address) {
     return ses.getIdentityVerificationAttributes({Identities: [address]}).promise()
         .then(response => {
             const address_verified = address in response.VerificationAttributes
-                && response.VerficationAttributes[address].VerificationStatus != 'Failed';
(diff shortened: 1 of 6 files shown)


lib/aws/ses-calls.js
```diff
@@ -0,0 +1,35 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+const AWS = require('aws-sdk');
+const winston = require('winston');
+
+function verifyEmailAddress(address) {
```
dsw88: I don't see a test on this function, could we get one added before I merge the PR?

brettbeatty: Okay, I added those tests.

PR #318, opened 2017-11-13
DynamoDB Autoscaling - Use DependsOn to avoid CloudWatch API Rate Limits

ThatJoeMoore:
Cloudwatch has very low API Rate Limits. Turns out, it's pretty easy to
hit that limit when setting up lots of application autoscaling rules in
Cloudformation (see https://forum.serverless.com/t/autoscaling-for-amazon-dynamodb/2071). So, by adding a DependsOn to all of our scaling targets,
we can avoid hitting that rate limit by forcing Cloudformation to only
make one call at a time, rather than parallelizing all of them.

In order to make that happen, I've also restructured the autoscaling
template.  It should produce the same output as before, with added
'DependsOn' attributes, but with a lot less duplication in the
Handlebars template itself. Plus, the new structure makes it a LOT
easier to compute the 'DependsOn' values.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,6 @@
   "author": "David Woodruff",
   "license": "Apache-2.0",
   "dependencies": {
-    "@types/uuid": "^3.4.3",
     "ajv": "^5.2.3",
     "ajv-errors": "^1.0.0",
     "archiver": "^1.3.0",
@@ -38,6 +37,7 @@
     "@types/mocha": "^2.2.44",
     "@types/node": "^8.0.47",
     "@types/sinon": "^2.3.7",
+    "@types/uuid": "^3.4.3",
     "@types/winston": "^2.3.7",
     "aws-sdk-mock": "^1.6.1",
     "chai": "^3.5.0",
@@ -52,8 +52,8 @@
     "run-sequence": "^2.2.0",
     "sinon": "^1.17.7",
     "sinon-chai": "^2.13.0",
+    "ts-node": "^3.3.0",
     "tslint": "^5.8.0",
-    "typescript": "^2.5.3",
-    "ts-node": "^3.3.0"
+    "typescript": "^2.5.3"
   }
 }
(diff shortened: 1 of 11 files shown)


dsw88:
The change looks great! I just had one thought about it: The index file is getting a little long at this point. What do you think about splitting the auto-scaling stuff out into its own file in this directory? It might help clarify the top-level deployer functionality to package that stuff away into its own module.

ThatJoeMoore:
Yeah, I like that.  Updated PR incoming.

ThatJoeMoore:
Alright, I've got the autoscaling config in its own (typescript!) file.

This commit also implicitly proposes a method for handling our
Typescript conversions, namely, creating a separate file (config-types.ts) containing only
the types used in configuring a given service. This allows us to split
some parts of a service implementation into separate files without
creating a circular dependency.

package.json
```diff
@@ -52,8 +52,8 @@
     "run-sequence": "^2.2.0",
     "sinon": "^1.17.7",
     "sinon-chai": "^2.13.0",
+    "ts-node": "^3.3.0",
     "tslint": "^5.8.0",
-    "typescript": "^2.5.3",
-    "ts-node": "^3.3.0"
+    "typescript": "^2.5.3"
```
ThatJoeMoore: Looks like the latest version of NPM changed their alphabetizing rules...

src/datatypes/service-context.ts
```diff
@@ -16,20 +16,21 @@
  */
 
 import { AccountConfig } from './account-config';
+import ServiceConfig from './service-config';
 
-export class ServiceContext {
+export class ServiceContext<Config extends ServiceConfig = any> {
```
ThatJoeMoore: This and the below changes let us optionally specify what type the params will be, but it leaves all current usages intact and unmodified. I really, really like having default values for types!

PR #321, opened 2017-12-05
Add support for DynamoDB TTL.

ThatJoeMoore:
See http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/TTL.html for details on this feature.

### DIFF
--- a/docs/supported-services/dynamodb.rst
+++ b/docs/supported-services/dynamodb.rst
@@ -42,6 +42,11 @@ Parameters
      - No
      - 5 for read and write
      - The ProvisionedThroughput element details how much provisioned IOPS you want on your table for reads and writes.
+   * - ttl_attribute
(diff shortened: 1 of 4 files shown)


dsw88:
Question about this PR: Do you need to configure the TTL attribute in the "Attributes" section of the CloudFormation template? I know it makes you do that when you configure the hash and range keys. Does it require you to explicitly configure it there for the TTL attribute as well?

dsw88:
@ThatJoeMoore This hasn't had any activity for a while, did you see my question a while ago?

ThatJoeMoore:
Okay, I've got this up-to-date with master.

I've done some testing, and DDB doesn't require you to list the TTL attribute. The way that you can set an item to not expire is by not setting a TTL attribute.

PR #322, opened 2017-12-05
Remove legacy injected env var format.

ThatJoeMoore:
Now, it's just the DEPENDENCY_NAME_VALUE_NAME format.

We've kept the legacy version around for a while; it's about time we
drop it. This is purely unselfish on my part and has no relation to
the fact that I just hit the 4k limit on total env var length on one of
my lambdas. Nope, definitely not.

Speaking of which, we may want to come up with a strategy for dealing
with the fact that many services have a limit on environment variables.


dsw88:
I agree, it's time to dump the old style. I'm pretty sure 95%+ of people have migrated to the new style at this point, because I submitted PRs to pretty much everyone who was using Handel at the time we made this change.

I also agree we need to figure out some strategy for some of these hard limits that we can run into like max env variable size or max cloudformation template size.

PR #334, opened 2017-12-18
added ability to sns to consume cloudwatch events

ndpete:
This probably isn't the cleanest way to do it but this is working.

SNS permissions are nearly identical to SQS permissions so I did a lot of copy and pasting from that and modifying where needed.

As another note I noticed some typos in the test descriptions for sqs and fixed that and added another test to sqs while I was at it.


src/aws/sns-calls.js
```diff
@@ -17,6 +17,103 @@
...
+        });
+}
+
+exports.addSnsPermissionIfNotExists = function (topicArn, sourceArn, policyStatement) {
+    return exports.getSnsPermission(topicArn, policyStatement)
+        .then(permission => {
+            if (!permission) {
+                return exports.addSnsPermission(topicArn, sourceArn, policyStatement)
+                    .then(() => {
+                        return exports.getSnsPermission(topicArn, policyStatement);
```
dsw88: Is the thing returned from addSnsPermission the same as what's returned from getSnsPermission? Often AWS will return the same object from a create that they would from a get, so if that's the case here then we can remove this call and just return what comes back from the create.

I have seen cases where AWS didn't do this however, forcing me to call get after a create like you're doing here.

ndpete: I believe they are different as there is no getSnsPermission. This function is a wrapper. It calls getTopicAttributes and then parses that for the PolicyDoc. This could be confusing too as there is a call with sns called addPermission it  is used for adding permissions across accounts. In this case we actually just call setTopicAttributes to add a policy doc. and getTopicAttributes to retrieve the policy doc.

dsw88: Oh yeah that rings a bell, sorry I had forgotten how the permission in SNS are part of the topic attributes.

dsw88:
Thanks for implementing this!

PR #350, opened 2018-01-10
this fixes alexa skills kit issue

### LINKED ISSUE #348: Duplicate alexa skills kit events
Every time handel deploy is run to update a lambda that has alexa skills kit as an event a duplicate event source is added as can be seen in the screenshot with the "(4)" by skills kit and the muliple entries when its clicked.

screen shot 2018-01-09 at 4 39 12 pm

ndpete:
This is a quick fix that resolves #348 

Not sure if this is the best way but it is functional.


dsw88:
I don't love the proposed fix, mostly because it injects knowledge of one of the services in the lambda-calls module, which I always try to avoid. It would be nice to see if we can do this some other way inside the service deployer itself before we make the choice to go down this route.

ndpete:
@dsw88 I rewrote the fix and have it incorporated in the lambda service now instead of lambda calls. I'd love your thoughts on the code.

dsw88:
@ndpete Unfortunately we did a major restructure of the project files because we had to add a second NPM module to the repository. This means the files you are trying to modify no longer exist in the location. It will probably be easiest for you to just want to take these changes and reapply them to what's currently on master in an updated PR.

ndpete:
@dsw88 The merge with all the changes was actually pretty simple. There was only one small conflict and I resolved that and pushed the changes back up. Let me know if this is good or not.

dsw88:
@ndpete I took a look at the PR (sorry it took so long!), and I think I found a pretty easy optimization that allows us to still using the existing code to add permissions for Alexa skill kit. Basically I just added a check for whether the sourceArn is defined to the part that checks whether it is equal to the policy. 

I pushed up a commit to this PR to show my proposed fix after looking at it. Can you take a look at it and tell me what you think about it?

ndpete:
Yeah I'll take a look at it. The issue was adding duplicate permissions not adding additionals. I'll take a look and run it.

ndpete:
I ran this through multiple tests and it works as expected. I feel silly I didn't  find the "easy" solution the first time around. Thanks @dsw88

dsw88:
Haha well I didn't see it either the first time, just saw it today when I looked again. I'll merge it!

Issue #332, opened 2017-12-15
Resource Tags


wiederhold:
Currently Handel is excellent at auto-injecting the following required tags:

  **name, app, env**

Unfortunately the following tags are totally optional:

  **team, data-sensitivity, platform, domain**

While all of these tags are currently required by policy, our mechanisms for enforcement aren't really effective. We can start shutting down non-compliant resources, but identifying who to notify of that action will be especially problematic without even a '**team**' tag.

* Recommend implementing some form of a **NetID** tag that corresponds with the person deploying resources. With that information we can track down the corresponding person most likely to be responsible/knowledgable about the deployment. This would be updated with each deployment.

The following tags remain largely unadopted for a variety of reasons:

  **data-sensitivity, platform, domain**

If we're going to require them by policy, we ought to provide some way to help folks meet that policy. As Handel gains traction as the first option for deployment, we could cover **A LOT** of ground by simply requiring these tags at deployment.

* Recommend implementing some form of deployment switch **-t** for the remaining tags **OR** integrating an interview process in into the **deployPhase switch** @ line 74

If this is what the Steering Committee intended with the 'enforced at deployment', I'm in vociferous support. If the underlying resources don't support tags, no worries. If Handel doesn't support the underlying service, same. With Handel support we'll make **SIGNIFICANT** tactical progress on this issue.

ThatJoeMoore:
Perhaps what we need is an equivalent to Cloudformation's stack tags - some top-level tags in the handelfile that get applied to all resources.

We wouldn't want to start enforcing strict BYUisms in Handel core, as it's been purposefully kept very generic, but perhaps we could add a list of required tags to the account config files and have Handel validate that they've been set? Then we get this kind of enforcement in a generic way.

wiederhold:
Sounds like an excellent strategy.

I totally get the open-source, share with the community thing--as long as the University's needs are top of mind as well.

Ultimately we could benefit from any form of personnel identification. BYUisms could be enforced post hoc with that information. And, that IS a generalizable issue. From a maintenance perspective, every CI in any medium to large size size organization requires an identifiable owner, a point of contact.

Currently we have too many resources deployed where our only clue about the stack are inferences from an app name or the id of the account it's deployed in. That is not sustainable.

ThatJoeMoore:
[...]

PR #362, opened 2018-01-25
Application-level tagging

ThatJoeMoore:
This PR will aim to make it easier to apply tags across all resources in a Handelfile.

For the BYU use case, this makes it easier to have consistent tagging with required tags like 'team' and 'data-sensitivity'.

This gets us part of the way towards solving the issues raised in #332.

### DIFF
--- a/src/common/deploy-phase-common.ts
+++ b/src/common/deploy-phase-common.ts
@@ -192,16 +192,17 @@ export function getResourceName(serviceContext: ServiceContext<ServiceConfig>) {
 }
 
 export function getTags(serviceContext: ServiceContext<ServiceConfig>): Tags {
-    let tags: Tags = {
-        app: serviceContext.appName,
-        env: serviceContext.environmentName
+    const serviceParams = serviceContext.params;
+
+    const handelTags: Tags = {
+             app: serviceContext.appName,
+             env: serviceContext.environmentName
     };
 
-    // Add user-defined tags if specified
-    const serviceParams = serviceContext.params;
-    if (serviceParams.tags) {
-        tags = Object.assign(tags, serviceParams.tags);
-    }
+    const appTags = serviceContext.tags;
 
-    return tags;
+    const serviceTags = serviceParams.tags;
+
+    // Service tags overwrite app tags, which overwrite Handel tags
+    return Object.assign({}, handelTags, appTags, serviceTags);
 }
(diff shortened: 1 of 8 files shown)


ThatJoeMoore:
So far, this is just the docs.  I wanted to get the feedback loop going while I start implementing.

Pinging @dsw88

dsw88:
This looks great! It's definitely a needed feature to reduce duplication for most tagging situations.

I do have one question that's not in the docs yet. When there is both a top-level tag and a resource-level tag with the same name, which one wins? I definitely prefer the resource-level tags overriding any top-level tags, and I imagine that is probably what you are expecting too.

ThatJoeMoore:
That's what I'd expect.  My implementation plan is to apply the application-level tags to the cloudformation stacks. The resource tags would remain on the individual resources, and in CFN, I'm pretty sure that resource tags override stack tags.

dsw88:
I've been going with the pattern that tags are applied to both the stack as well as the applicable resources inside the stack. I was assuming global tags would follow this same pattern where they are applied to every stack as well as the resources inside them. What do you think about continuing that pattern?

ThatJoeMoore:
This is now ready for review.  I ended up piggybacking on the existing tag code, so I believe that the tags will get propagated everywhere they were before, just like 'app' and 'env'.

I'm not sure we need to propagate all tags to all resources inside of a CFN stack, because CFN auto-tags everything in the stack with the stack tags (if not overwritten by the resource tags).

ThatJoeMoore:
Fun fact: with this PR, we have an even 500 test cases in our suite ✋ 💯

src/common/deploy-phase-common.ts
```diff
@@ -192,16 +192,17 @@ export function getResourceName(serviceContext: ServiceContext<ServiceConfig>) {
...
-    if (serviceParams.tags) {
-        tags = Object.assign(tags, serviceParams.tags);
-    }
+    const appTags = serviceContext.tags;
+
+    const serviceTags = serviceParams.tags;
 
-    return tags;
+    // Service tags overwrite app tags, which overwrite Handel tags
+    return Object.assign({}, handelTags, appTags, serviceTags);
```
dsw88: I was about to ask why you were adding these when you haven't checked that they exist, but then I went and tried it out and it turns out Object.assign handles undefined values without a problem. You learn something new every day!

ThatJoeMoore: Yup!  That was my TIL from a couple weeks ago :D

dsw88:
Thanks for implementing this @ThatJoeMoore!

PR #364, opened 2018-01-25
Add required tags to account config

ThatJoeMoore:
Per the discussion in #332, this PR will allow us to add a list of required tags to an account config file. This list will then be enforced during `handel check` and `handel deploy`.

Note: This won't take effect immediately when merged. We'd have to update each account config file, then have all deployers download the new account config file.

Side note: should we explore keeping the account config files in a specially-named S3 bucket or Parameter Store value? Then we could roll out account config file changes more easily.

### DIFF
--- a/src/services/apiaccess/index.ts
+++ b/src/services/apiaccess/index.ts
@@ -77,3 +77,5 @@ export const producedDeployOutputTypes = [
 ];
 
 export const consumedDeployOutputTypes = [];
+
+export const supportsTagging = true;
--- a/src/services/apiaccess/index.ts
+++ b/src/services/apiaccess/index.ts
@@ -78,4 +78,4 @@ export const producedDeployOutputTypes = [
 
 export const consumedDeployOutputTypes = [];
 
-export const supportsTagging = true;
+export const supportsTagging = false;
--- a/src/services/cloudwatchevent/index.ts
+++ b/src/services/cloudwatchevent/index.ts
@@ -22,8 +22,16 @@ import * as deletePhasesCommon from '../../common/delete-phases-common';
 import * as deployPhaseCommon from '../../common/deploy-phase-common';
 import * as handlebarsUtils from '../../common/handlebars-utils';
 import * as produceEventsPhaseCommon from '../../common/produce-events-phase-common';
-import { AccountConfig, DeployContext, PreDeployContext, ProduceEventsContext, ServiceConfig, ServiceContext, ServiceEventConsumer, UnDeployContext } from '../../datatypes';
-import { CloudWatchEventsConfig, CloudWatchEventsServiceEventConsumer } from './config-types';
+import {getTags} from '../../common/tagging-common';
+import {
+    DeployContext,
+    PreDeployContext,
+    ProduceEventsContext,
+    ServiceConfig,
+    ServiceContext,
+    UnDeployContext
+} from '../../datatypes';
+import {CloudWatchEventsConfig, CloudWatchEventsServiceEventConsumer} from './config-types';
 
 const SERVICE_NAME = 'CloudWatch Events';
 
@@ -85,7 +93,7 @@ export async function deploy(ownServiceContext: ServiceContext<CloudWatchEventsC
     winston.info(`${SERVICE_NAME} - Deploying event rule ${stackName}`);
 
     const eventRuleTemplate = await getCompiledEventRuleTemplate(stackName, ownServiceContext);
-    const stackTags = deployPhaseCommon.getTags(ownServiceContext);
+    const stackTags = getTags(ownServiceContext);
     const deployedStack = await deployPhaseCommon.deployCloudFormationStack(stackName, eventRuleTemplate, [], true, SERVICE_NAME, stackTags);
     winston.info(`${SERVICE_NAME} - Finished deploying event rule ${stackName}`);
     return getDeployContext(ownServiceContext, deployedStack);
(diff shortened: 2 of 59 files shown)


ThatJoeMoore:
@dsw88 ready for review.

ThatJoeMoore:
One sad thing - this PR ruins our beautiful test count of 500.  Now it's 501 😭

dsw88:
Oh you're right! You'd better delete one of our tests before we merge so we can get back to 500. 😄 Haha jk, I'm reviewing now.

src/services/apiaccess/index.ts
```diff
@@ -77,3 +77,5 @@ export const producedDeployOutputTypes = [
 ];
 
 export const consumedDeployOutputTypes = [];
+
+export const supportsTagging = true;
```
dsw88: I think this one should be set to false right?

src/services/cloudwatchevent/index.ts
```diff
@@ -145,3 +153,5 @@ export const producedEventsSupportedServices = [
 export const producedDeployOutputTypes = [];
 
 export const consumedDeployOutputTypes = [];
+
+export const supportsTagging = true;
```
dsw88: I think the Handel docs don't show that the CloudWatch Events service supports Tags (the rules don't, but we do tag the stack here), can we add that to the docs as part of this change?

src/services/iot/index.ts
```diff
@@ -159,3 +167,5 @@ export const producedEventsSupportedServices = [
 export const producedDeployOutputTypes = [];
 
 export const consumedDeployOutputTypes = [];
+
+export const supportsTagging = true;
```
dsw88: The Handel docs also don't show that Tags are supported, but we apparently do add tags to the stack. Can we update those docs as part of this PR?

src/services/kms/index.ts
```diff
@@ -123,3 +124,5 @@ export const producedDeployOutputTypes = [
 ];
 
 export const consumedDeployOutputTypes = [];
+
+export const supportsTagging = true;
```
dsw88: The Handel docs also don't show that Tags are supported for KMS, but we apparently do add tags to the stack. Can we update those docs as part of this PR?

src/services/ses/index.ts
```diff
@@ -84,3 +84,5 @@ export const producedDeployOutputTypes = [
 ];
 
 export const consumedDeployOutputTypes = [];
+
+export const supportsTagging = true;
```
dsw88: I think this one should be set to false right?

src/services/sns/index.ts
```diff
@@ -182,3 +191,5 @@ export const producedDeployOutputTypes = [
 export const consumedDeployOutputTypes = [
     'cloudwatchevent'
 ];
+
+export const supportsTagging = true;
```
dsw88: The Handel docs also don't show that Tags are supported for KMS, but we apparently do add tags to the stack. Can we update those docs as part of this PR?

src/services/sqs/index.ts
```diff
@@ -230,3 +238,5 @@ export const producedDeployOutputTypes = [
 ];
 
 export const consumedDeployOutputTypes = [];
+
+export const supportsTagging = true;
```
dsw88: The Handel docs also don't show that Tags are supported for KMS, but we apparently do add tags to the stack. Can we update those docs as part of this PR?

dsw88:
The PR looks great! Just a few comments, mostly about the fact that historically I haven't shown the "tags" parameter on things like SQS that don't support tags for the queue itself, but we are tagging the stacks at least and any of the resources that do support it. I'm thinking we should update the docs to show the tags element for those, what do you think?

wiederhold:
> Per the discussion in #332, this PR will allow us to add a list of required tags to an account config file. This list will then be enforced during handel check and handel deploy.

Note: This won't take effect immediately when merged. We'd have to update each account config file, then have all deployers download the new account config file.

Side note: should we explore keeping the account config files in a specially-named S3 bucket or Parameter Store value? Then we could roll out account config file changes more easily.

☝️ Yes!!! I also REALLY like the idea of enforcement for handel configs at deploy. By putting them in a central location we can get to an ideal scenario where any organization (not just BYU) could enforce there own set of rules/guidelines at deployment for an entire stack. This is WAY better than AWS Config in its present state, and more to the point, gets the config rules closer to the dev pipeline than any other mechanism anyone's suggested. HUGE thumbs up... this is IMO brilliant!

ThatJoeMoore:
Adding tags to the tagless services works for me.  We should probably note that, at this time, those resources don't support tagging, so we'll just be tagging the cloudformation.  But if AWS ever gets their act together, the tags should get automatically deployed the next time the CFN stack gets updated, so that could be really nice.

ThatJoeMoore:
I think I'm going to go ahead and restructure some of our tagging documentation. Every service's tag schema is the same, so instead of documenting it in every single service, I'm going to have them all link to the main tagging document.  I think that'll make all of this much easier to keep clean and up-to-date.

dsw88:
Great idea, yeah there's a lot of repeated text for tags on every service page, it would definitely be nice to consolidate it into a single location.

ThatJoeMoore:
Alright, I think I've got all the docs nice and pretty.

dsw88:
Awesome looks awesome, thanks for implementing this!

PR #368, opened 2018-01-26
Implement CLI Tag Overrides

ThatJoeMoore:
Yet another thing related to #332.

Allows the user to pass a -t flag to `handel deploy` with a list of
key-value pairs. These will augment or override any tags that are
specified in the handelfile.

### DIFF
--- a/docs/getting-started/cli-reference.rst
+++ b/docs/getting-started/cli-reference.rst
@@ -1,0 +1,101 @@
+.. _cli-reference:
+
+CLI Reference
+=============
+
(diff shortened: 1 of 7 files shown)


docs/getting-started/cli-reference.rst
```diff
@@ -0,0 +1,101 @@
+.. _cli-reference:
+
+CLI Reference
```
dsw88: Thanks for adding this to the docs site, it's definitely helpful to have on the site docs as well as on the CLI itself.

PR #369, opened 2018-01-26
Tag handel account-wide resources

### LINKED ISSUE #360: Handel needs to apply tags to the lambdas that are automatically deployed under the covers
Handel deploys several lambdas in different scenarios to manage different resources or provide additional functionality. These lambdas include (but are not limited to):
- `HandelEcsAutoScalingLambda`
- `HandelEcsDrainingLambda`
- `HandelSlackNotify`
- etc

We need to have a way to apply tags to these Lambdas in order to meet enterprise requirements.

ThatJoeMoore:
Fixes #360.

Adds an optional field to the account config file, handel_resource_tags,
which contains tags that Handel will apply to any non-application,
account-wide resources that Handel may create.

This change will not be able to modify any resources that already exist.
Those will either have to be deleted and re-created, or one can manually
add the tags to the cloudformation templates that Handel creates to add
the tags.

### DIFF
--- a/src/aws/aws-tags.ts
+++ b/src/aws/aws-tags.ts
@@ -1,0 +1,18 @@
+import {Tags} from '../datatypes';
+
+export interface GenericAWSTag {
+    Key: string;
+    Value: string;
+}
+
+export type GenericAWSTagSet = GenericAWSTag[];
+
+export function toAWSTagStyle(tags: Tags): GenericAWSTagSet {
+    return Object.getOwnPropertyNames(tags)
+        .map(name => {
+            return {
+                Key: name,
+                Value: tags[name]
+            };
+        });
+}
--- a/src/common/pre-deploy-phase-common.ts
+++ b/src/common/pre-deploy-phase-common.ts
@@ -17,11 +17,12 @@
 import * as winston from 'winston';
 import * as cloudformationCalls from '../aws/cloudformation-calls';
 import * as ec2Calls from '../aws/ec2-calls';
-import { AccountConfig, PreDeployContext, ServiceConfig, ServiceContext } from '../datatypes';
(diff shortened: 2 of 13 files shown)


src/aws/aws-tags.ts
```diff
@@ -0,0 +1,18 @@
+import {Tags} from '../datatypes';
+
+export interface GenericAWSTag {
+    Key: string;
+    Value: string;
+}
+
+export type GenericAWSTagSet = GenericAWSTag[];
+
+export function toAWSTagStyle(tags: Tags): GenericAWSTagSet {
```
dsw88: Thanks for adding this module, it was needed!

src/common/pre-deploy-phase-common.ts
```diff
@@ -17,11 +17,12 @@
...
 import * as cloudformationCalls from '../aws/cloudformation-calls';
 import * as ec2Calls from '../aws/ec2-calls';
-import { AccountConfig, PreDeployContext, ServiceConfig, ServiceContext } from '../datatypes';
+import {AccountConfig, PreDeployContext, ServiceConfig, ServiceContext, Tags} from '../datatypes';
 import * as deployPhaseCommon from './deploy-phase-common';
 import * as handlebarsUtils from './handlebars-utils';
+import {getTags} from './tagging-common';
 
-async function createSecurityGroupForService(stackName: string, sshBastionIngressPort: number | null, accountConfig: AccountConfig) {
+async function createSecurityGroupForService(stackName: string, sshBastionIngressPort: number | null, accountConfig: AccountConfig, tags: Tags) {
```
dsw88: Wow thanks for catching this, I didn't even realize our security group stacks weren't being tagged!

ThatJoeMoore: You can thank IntelliJ's Find Usages inspection :grin: I just ran a find usage on the create stack call and went from there.

PR #372, opened 2018-01-30
Add extension API Types

ThatJoeMoore:
This depends on #371 being merged.

Adds an extension-api package which contains all of the data types needed by the extensions API. This is a subset of the types defined in handel/src/datatypes/index.

The old datatypes file now extends or implements the new one in all cases, which allows us to add internal extensions to anything we need to while enforcing compatibility with the extensions API.

Some thoughts to consider while reviewing:

* Are there any fields in the interfaces that we don't need to be publicly available?
* For the service response types, should we keep them as classes? Right now, they're interfaces, but providing the class version might be nice for people who have to construct them to return from the methods.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -2,8 +2,7 @@
   "lerna": "2.8.0",
   "packages": [
     "handel",
-    "extension-api",
-    "official-extensions"
+    "extension-api"
   ],
   "version": "0.23.1"
 }
--- a/lerna.json
+++ b/lerna.json
@@ -1,7 +1,9 @@
 {
   "lerna": "2.8.0",
   "packages": [
-    "handel"
+    "handel",
+    "extension-api",
+    "official-extensions"
   ],
   "version": "0.23.1"
 }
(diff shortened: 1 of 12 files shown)


lerna.json
```diff
@@ -1,7 +1,9 @@
 {
   "lerna": "2.8.0",
   "packages": [
-    "handel"
+    "handel",
+    "extension-api",
+    "official-extensions"
```
dsw88: This looks like a placeholder for something you had in mind. Are you thinking to split out what's currently in the services/ directory in Handel to be modules inside this directory, or were you thinking this would be for something else?

dsw88:
You still have [BLOCKED] on the PR title, but it looks like you've sorted out this PR with the changes applied in #371 at this point. Do you still have more to do on this, or did you just forget to remove that from the title?

lerna.json
ThatJoeMoore: Huh. I had meant to remove that; I was just seeing how Lerna handled paths that don't exist (spoiler alert: It just ignores them!).

I had thought of adding 'official' extensions at one point to add some interesting services, but then I realized that we should probably just add anything like that into the core library 🙃

ThatJoeMoore:
It should all be good to go now. I hadn't removed the 'blocked' because, as it turns out, you can't edit a PR from GH's mobile interface 🙁

PR #373, opened 2018-02-02
Fargate polishing

### LINKED ISSUE #325: Fargate: Flush out the check method
The check method has some more work to do.  

We need to validate the memory and cpu.

dsw88:
Resolves #325, #326, #327, #328, and #329


dsw88:
The unit tests are failing in Circle, but I can't replicate them on any other machine. I'm merging this for now, and will look at the failing tests (it could be a caching issue in Circle)

PR #385, opened 2018-02-23
API Gateway Custom Domain Support

### LINKED ISSUE #347: API Gateway - Support custom domains and BasePath mappings
APIGW allows you to setup up custom domain name DNS entry/certificate pairs and create a mapping from them to an API/stage pair.  One can associate multiple APIs/stages with a single custom domain name using a base path - IE, you could route to API A / Stage Prod with `https://myapi.example.com/api-a-prod` and API B / Stage Beta with `https://myapi.example.com/api-b-beta`, etc - all off of the same domain name.

I'd like us to support setting up these kinds of mappings.

So, the dependency graph between the moving parts here is like this:

domain

The most straightforward way to set this up (though it may not be the easiest way due to the code structure) is to only allow one domain name and base path mapping per API Gateway service.  So, it would look like this:

```yaml
  type: apigateway
  proxy: ...
  base_path:
    dns_name: myapi.example.com
    path: /mypath # default: /
    certificate_id: arn:a:b:c:d
```

We could also easily allow mapping multiple base paths per api by making that an array.

Another option would be to have a new service like `apigateway_custom_domain` which, combined with a `base_path` attribute on the gateway, would allow us to map multiple different APIs onto the same domain:

```yaml
custom-domain:
  type: apigateway_custom_domain
  dns_name: myapi.example.com
  certificate_id: arn:a:b:c:d

api-a:
  type: apigateway
  proxy: ...
  base_paths:
  - domain: custom-domain
     path: /api-a

api-b:
  type: apigateway
  proxy: ...
  base_paths:
  - domain: custom-domain
     path: /api-b

```

We could also allow both - easily configure the custom domain using the dns_name and certificate_id, or referencing a shared one using the `domain` or `custom_domain` or something.

I can do the implementation (as I need it soon-ish), but what do you think about the different configuration options?

ThatJoeMoore:
This resolves #347.

This implements the simplest-possible version of custom domain names.  It only allows for a custom domain name to be used by a single API Gateway, and doesn't allow for path mappings (everything is at /).  If we want to enhance that later, we can.

I also factored some common checking logic between the two API gateway deployers out into the main service file.

### DIFF
--- a/docs/supported-services/apigateway.rst
+++ b/docs/supported-services/apigateway.rst
@@ -10,6 +10,11 @@ No Authorizer Lambdas
 ~~~~~~~~~~~~~~~~~~~~~
 This service doesn't yet support specifying authorizer lambdas.
 
+No Regional Endpoints
+~~~~~~~~~~~~~~~~~~~~~
+
+This service currently supports only edge-optimized API Gateways.
+
 Parameters
 ----------
 
@@ -79,7 +84,7 @@ The Custom Domains section is defined by the following schema:
 
     custom_domains:
     - dns_name: <string> # The DNS name for the API Gateway. Must be a valid DNS name.
-      https_certificate: <arn> # The Amazon Certificate Manager certificate to use. This certificate must be in the same region as the API Gateway instance.
+      https_certificate: <arn> # The Amazon Certificate Manager certificate to use. This certificate must be in the us-east-1 region.
 
 See :ref:`route53zone-records` for more information on how DNS records will be created.
 
--- a/docs/supported-services/apigateway.rst
+++ b/docs/supported-services/apigateway.rst
@@ -51,11 +51,37 @@ Parameters
      - No
      - false
      - If true, your Lambdas will be deployed inside your account's VPC.
+   * - custom_domains
+     - Array of :ref:`apigateway-custom-domains`
+     - No
+     -
+     - An array of custom domains to map to this API Gateway instance.
    * - tags
      - :ref:`tagging-resources`
      - No
      - 
      - Any tags you want to apply to your API Gateway app.
+
+.. _apigateway-custom-domains:
+
+Custom Domain Mappings
+~~~~~~~~~~~~~~~~~~~~~~
+.. NOTE::
+
+    This service does not currently support sharing custom domains between API Gateway instances using Base Path Mappings.
+    At this time, you can only map one API Gateway to one custom domain, with no path mapping.
+
+API Gateway allows for mapping gateways to one or more custom domains. These custom domains are always served via HTTPS.
+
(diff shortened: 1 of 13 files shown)


docs/supported-services/apigateway.rst
```diff
@@ -51,12 +56,38 @@ Parameters
...
+
+API Gateway allows for mapping gateways to one or more custom domains. These custom domains are always served via HTTPS.
+
+The Custom Domains section is defined by the following schema:
+
+.. code-block:: yaml
+
+    custom_domains:
+    - dns_name: <string> # The DNS name for the API Gateway. Must be a valid DNS name.
+      https_certificate: <arn> # The Amazon Certificate Manager certificate to use. This certificate must be in the us-east-1 region.
```
dsw88: This is an interesting comment about the certificate needing to be in the us-east-1 region, is this an AWS limitation? What happens if I want to use custom domains but am deploying to another region like us-west-2?

ThatJoeMoore: It's a limitation of how they implement custom domains for API Gateway.  The dirty little secret is that API Gateway is essentially a fancy multi-tenant Cloudfront distribution. When you create a custom domain, it creates a new Cloudfront distribution under the hood. Cloudfront requires that all resources that it uses (certs, Lambda@Edge) be in us-east-1.

Lambda did recently add a new option - "Endpoint Configuration." The default value is 'edge', which matches the legacy way of using API Gateway and uses Cloudfront for everything. There's a new value, 'regional', which sets up a region-specific API endpoint that doesn't use Cloudfront and keeps the certs in the specific region. I decided that adding support for that was beyond the scope of this PR.  It wouldn't be hard, though, as we could basically just have a config like this:

```yml
  endpoint_configuration:
    types:
    - regional
    - edge
```

The two types of endpoint apparently aren't mutually exclusive.  We'd then have to add some extra logic to what I've added in this PR to handle that case and configure thing properly, but I don't think it'd be too hard.  The only real difficulty there would be that we need to have each `custom_domain` know which kind of endpoint it's supposed to point at.

Perhaps a better config for that would be to ditch the above and do this:

```yml
  custom_domains:
  - dns_name: regional.mycompany.com
     https_certificate: arn:acm:us-west-2:something
     type: 'regional'
  - dns_name: edge.mycompany.com
     https_certificate: arn:acm:us-east-1:something
     type: 'edge'
```
.
Then, based on the values in the custom domains, we set the values of 'EndpointConfiguration.Type', so the user doesn't have to worry about them.  We'd default to just 'Edge'.

FYI, here's the docs for the Cloudformation involved:

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-apigateway-restapi-endpointconfiguration.html

PR #390, opened 2018-02-26
Remove SetIPV6 Function from s3staticsite

ThatJoeMoore:
Cloudformation finally supports setting IPV6Enabled on cloudfront distributions, so we don't need our custom resource anymore to do it.  So, I got rid of it 🗑 💥 .

I also did a little bit of asyncawaiting in the function where all the magic used to happen.


dsw88:
Hooray, I love deleting code, thanks for doing this!

PR #391, opened 2018-02-26
Fix certificate ARNs for APIGW custom domains

ThatJoeMoore:
Right now, everything besides API gateway that uses ACM certs accepts the
certificate ID. APIGW doesn't. This makes APIGW work with both ARNs and
certificate IDs.  In the future, we may want to make sure that all
other services also allow both.


dsw88:
Hmm I don't know that I love having people have to specify full ARNs in their Handel files. Could we still have them provide just the ID, but construct the ARN internally from that?

ThatJoeMoore:
That's actually what this is fixing.  When I initially wrote the API gateway custom domain stuff, I accidentally made it require the full ARN.  Now, it'll construct it from the ID, but still allows you to pass the full ARN if you really really want to.

dsw88:
I guess I'm ok with supporting an ARN too.

PR #395, opened 2018-03-12
implementing stepfunctions

brettbeatty:
I think I have it done. Let me know what you think, especially when it comes to the documentation. I wasn't too sure how to word the 'State Machine Definition' section.


handel/src/services/stepfunctions/index.ts
```diff
@@ -0,0 +1,138 @@
...
+    winston.info(`${SERVICE_NAME} - Executing Deploy on '${stackName}'`);
+    const compiledStepFunctionsTemplate = await getCompiledStepFunctionsTemplate(stackName, ownServiceContext, dependenciesDeployContexts);
+    const stackTags = getTags(ownServiceContext);
+    const deployedStack = await deployPhaseCommon.deployCloudFormationStack(stackName, compiledStepFunctionsTemplate, [], true, SERVICE_NAME, 30, stackTags);
+    winston.info(`${SERVICE_NAME} - Finished deploying '${stackName}'`);
+    return getDeployContext(ownServiceContext, deployedStack);
+}
+
+export async function unDeploy(ownServiceContext: ServiceContext<StepFunctionsConfig>): Promise<UnDeployContext> {
+    await iamCalls.detachPoliciesFromRole(deployPhaseCommon.getResourceName(ownServiceContext));
```
dsw88: I don't think you need to do this here, since the role and policies are contained inside the CloudFormation template that will be deleted on the next line. Am I missing something though?

handel/src/services/stepfunctions/index.ts
```diff
@@ -0,0 +1,138 @@
...
+    });
+
+    // Inject env vars
+    deployContext.addEnvironmentVariables(deployPhaseCommon.getInjectedEnvVarsFor(serviceContext, {
+        STATE_MACHINE_ARN: stateMachineArn,
+        STATE_MACHINE_NAME: stateMachineName
+    }));
+
+    // Inject event outputs
+    deployContext.eventOutputs.stateMachineArn = stateMachineArn;
```
dsw88: You can probably remove these few lines setting eventOutputs since this service won't produce or consume events from other services.

handel/src/services/stepfunctions/index.ts
```diff
@@ -0,0 +1,138 @@
...
+    const dependencyArns: Map<string, string> = new Map();
+    // Map service name to ARN
+    for (const context of dependenciesDeployContexts) {
+        dependencyArns.set(context.serviceName, context.eventOutputs.lambdaArn);
+    }
+    // Change 'resource' in each state from service name to ARN
+    for (const state_name in definitionFile.States) {
+        const state = definitionFile.States[state_name];
+        if (definitionFile.States.hasOwnProperty(state_name) && 'Resource' in state) {
+            state.Resource = dependencyArns.get(state.Resource);
```
dsw88: Could we add something in the check() method to validate the the lambdas named in the state file actually exist in the Handel file? Right now if you specify a service in your state file that doesn't exist in your Handel file, it will likely fail with a cryptic error for the user.

docs/supported-services/stepfunctions.rst
```diff
@@ -0,0 +1,106 @@
...
+     - :ref:`stepfunctions-definition`
+     - Yes
+     -
+     - Path to file containing state machine definition.
+
+.. _stepfunctions-definition:
+
+State Machine Definition
+~~~~~~~~~~~~~~~~~~~~~~~~
+For the most part, the definition file you provide in the *definition* section is in `Amazon States Language <https://docs.aws.amazon.com/step-functions/latest/dg/concepts-amazon-states-language.html>`_. Instead of providing an ARN in the 'Resource' field of a state, however, one should give the service name from the Handel file. For convenience, Handel supports both JSON and YAML formats for the definition file, where pure States Language is based on JSON alone.
```
dsw88: Can we make the last sentence starting with "For convenience" a Note box so that it pops out a bit more?

Also, I might recommend taking the example state_machine.yml file in the Example Handel File section and put it in this section instead, since this section is helping you figure out what needs to be provided for the *definition* parameter. Then you could put a comment in the example Handel file on the *definition* line that refers them to this section if they have questions about what to put in that file.

dsw88:
Ok @brettbeatty I reviewed this PR and put some comments in for you to review.

brettbeatty:
Okay I think I addressed all of those issues

dsw88:
Thanks so much for implementing this @brettbeatty! I've been thinking about using step functions myself on occasion, so I'm happy to have a service for it in Handel.

Issue #302, opened 2017-10-24
Implement external plugin system.


dsw88:
People would like Handel services for things like MongoDB and RabbitMQ, which AWS doesn't provide native services for.

Handel should focus on the core AWS services, but it should offer a plugin model that allows people to define service provisioners that can be imported into the deployment lifecycle.

ThatJoeMoore:
I've been thinking about how to implement this.

The code samples given below may conflict with each other - I'm basically walking through my thought/design process, so each is an evolution on the previous in order to solve the new problem.

**tl;dr:** While a super-basic plugins implementation would be easy to add, there are some complicated things we should at least think about.

# The plugins used in a handel stack must be explicitly declared

This helps make the handelfile more understandable, as one can see at a glance which plugins are being used.  It also solves a host of problems related to service names.

## What it looks like

Have a `plugins` section in `handel.yml`:

```yaml
version: 1
plugins:
- foobar-handel-plugin
```

## Alternatives

We could implement things Yeoman-style and require plugins to have a certain pattern to their names, as well as declaring a certain keyword in their package.json.  So, for example, a `foobar` plugin would have a (minimal) package.json like:

```json
{
  "name": "handel-foobar-plugin",
  "version": "1.0.0",
  "keywords": ["handel-plugin"]
}
```

This has the downside of making Handel more dependent on the user's local configuration, and making it harder to know which plugins must be installed for a given Handelfile to run.  It might, in fact, make it very difficult for something like Handel-Codepipeline to work properly with plugins, as it would have to somehow discover which plugins are being used.

# We should have the user namespace the plugins they use

If we force the user to define a namespace for each plugin they use, we automatically avoid service name conflicts. By making the user define the namespace, we can make some things more convenient for both them and us. They'll always have a simple name for each plugin, and we don't have to worry about plugins declaring conflicting 'short' namespaces (I'll explain this below).

## What it looks like

`handel.yml`
```yaml
version: 1

plugins:
  foobar: foobar-handel-plugin

environments:
  dev:
    foobar:
      type: foobar::foo
      foo: 'bar'
```

## Alternatives

There are two alternatives I can think of: using the NPM package name as the namespace and having each plugin declare a namespace for itself.

[...]

PR #406, opened 2018-03-30
Handel Extensions Episode I: The Phantom ServiceRegistry

ThatJoeMoore:
Supports #302 , but doesn't fix it.

This PR doesn't actually add any features to Handel - yet.  However, it adds the concept of a 'Service Registry', and forces everything to use that instead of `util.getServiceDeployers()`.

The Service Registry knows how to load extensions from a list of extension definitions, but doesn't know how to download/install extensions or read a list of them from the handelfile.  That'll come in a later PR.  Like I said, this commit changes nothing about handel files or the public API, just the internal APIs that we use.

I'm going to go through the commit and make comments on some key changes.


extension-api/extension-api.ts
```diff
@@ -77,13 +77,21 @@ export interface AccountConfig {
 /***********************************
  * Types for the context objects used by service deployers
  ***********************************/
+
+export interface ServiceRegistry {
+    findDeployerFor(prefix: string, name: string): ServiceDeployer;
+    hasService(prefix: string, name: string): boolean;
+    validPrefixes(): Set<string>;
```
ThatJoeMoore: Hmm.  I'm gonna actually push a commit that makes these names better.  Gimme a few minutes.

extension-api/package.json
```diff
@@ -10,11 +10,12 @@
   },
   "author": "David Woodruff",
   "license": "Apache-2.0",
-  "dependencies": {
+  "peerDependencies": {
     "aws-sdk": "^2.188.0"
```
ThatJoeMoore: By moving aws-sdk to a peer dependency, we can avoid having a million versions of aws-sdk running around inside of nested node_modules, which'll speed up our installs and startups. By like, a lot, because aws-sdk is HEAVY.

handel/gulpfile.js
```diff
@@ -5,7 +5,6 @@ const fs = require('fs-extra');
 const path = require('path');
 const runSeq = require('run-sequence');
 const mergeStream = require('merge-stream');
-const mocha = require('gulp-spawn-mocha');
```
ThatJoeMoore: This wasn't actually getting used and was giving me console warnings, so I killed it 💣

extension-api/package.json
```diff
@@ -10,11 +10,12 @@
...
   },
   "devDependencies": {
     "@types/node": "^9.4.0",
+    "aws-sdk": "^2.188.0",
```
ThatJoeMoore: ...and to make Typescript happy, we have to install it as a dev dependency too 🤷‍♂️

handel/package.json
```diff
@@ -18,7 +18,8 @@
     "ajv-errors": "^1.0.0",
     "archiver": "^1.3.0",
     "aws-sdk": "^2.169.0",
-    "fs-extra": "^4.0.2",
+    "common-tags": "^1.7.2",
```
ThatJoeMoore: This package gives us some nice utilities around tagged template literals, which can help us have nice pretty error messages in the extensions API (since I think that may get confusing to people if we don't have friendly error messages).

handel/src/datatypes/index.ts
```diff
@@ -34,28 +34,15 @@ export interface AccountConfig extends api.AccountConfig {
...
-                accountConfig: AccountConfig,
-                tags: Tags = {}) {
-            this.appName = appName;
-            this.environmentName = environmentName;
-            this.serviceName = serviceName;
-            this.serviceType = serviceType;
-            this.params = params;
-            this.accountConfig = accountConfig;
-            this.tags = tags;
+    constructor(public appName: string,
```
dsw88: Didn't even know you could do this. :)

handel/test/service-registry/fake-service-registry.ts
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+import {ServiceDeployer, ServiceRegistry} from 'handel-extension-api';
+import {DEFAULT_EXTENSION_PREFIX} from '../../src/service-registry';
+
+export interface FakeRegistryInfo {
+    [key: string]: Partial<ServiceDeployer>;
+}
+
+export default class FakeServiceRegistry implements ServiceRegistry {
```
ThatJoeMoore: Sooo much easier to just have a fake object than to use Sinon for this.

handel/src/datatypes/index.ts
ThatJoeMoore: Right!?!

handel/tsconfig.json
```diff
@@ -1,7 +1,8 @@
 {
   "extends": "../tsconfig",
   "compilerOptions": {
-    "allowJs": true
+    "allowJs": true,
+    "rootDir": "./"
```
dsw88: I looked at the tsconfig docs for this, but I'm still a little unclear on what it does. Can you tell me how this changes the compilation?

ThatJoeMoore: Among other things, it disallows relative paths that leave that directory, which is good because my IDE kept wanting to import the extension api from relative paths instead of from the node package.

ThatJoeMoore:
Per a face-to-face conversation with @dsw88 , I've removed ServiceRegistry from ServiceContext and replaced it with a ServiceInfo object.  That'll allow services to see information about their dependencies without having the ability to do unspeakable things with them.

dsw88:
This is looking great so far!

PR #408, opened 2018-04-04
Handel Extensions Episode II: Attack of the CLI

ThatJoeMoore:
This rewrites the CLI interface to use commander.js, which gets rid of a bunch of oddness.  It also now passes the options from the CLI to the Environment Context so we can do things like using CLI feature flags.

### DIFF
--- a/handel/package.json
+++ b/handel/package.json
@@ -5,7 +5,7 @@
   "main": "./bin/handel",
   "scripts": {
     "lint": "tslint -p tsconfig.json -t stylish",
-    "build": "tsc",
+    "build": "gulp build",
     "test": "./node_modules/mocha/bin/mocha -r ts-node/register 'test/**/*.ts'"
   },
   "bin": {
@@ -18,6 +18,8 @@
     "ajv-errors": "^1.0.0",
     "archiver": "^1.3.0",
     "aws-sdk": "^2.169.0",
+    "change-case": "^3.0.2",
+    "commander": "^2.15.1",
     "common-tags": "^1.7.2",
     "fs-extra": "^5.0.0",
     "handel-extension-api": "^0.25.1",
@@ -28,6 +30,7 @@
     "minimist": "^1.2.0",
     "ncp": "^2.0.0",
     "pascal-case": "^2.0.1",
+    "pjson": "^1.0.9",
     "tmp": "0.0.33",
     "uuid": "^3.0.1",
     "winston": "^2.3.1"
--- a/handel/package.json
+++ b/handel/package.json
@@ -27,7 +27,6 @@
     "inquirer": "^3.3.0",
     "js-yaml": "3.7.0",
     "lodash": "^4.17.4",
-    "minimist": "^1.2.0",
     "ncp": "^2.0.0",
     "pascal-case": "^2.0.1",
     "pjson": "^1.0.9",
--- a/handel/src/cli/index.ts
+++ b/handel/src/cli/index.ts
@@ -16,25 +16,34 @@
  */
 import * as AWS from 'aws-sdk';
 import * as fs from 'fs';
-import {ServiceRegistry} from 'handel-extension-api';
+import { ServiceRegistry } from 'handel-extension-api';
 import * as inquirer from 'inquirer';
 import * as yaml from 'js-yaml';
+import * as _ from 'lodash';
 import * as winston from 'winston';
 import config from '../account-config/account-config';
 import * as stsCalls from '../aws/sts-calls';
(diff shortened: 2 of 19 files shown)


handel/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "./bin/handel",
   "scripts": {
     "lint": "tslint -p tsconfig.json -t stylish",
-    "build": "tsc",
+    "build": "gulp build",
```
ThatJoeMoore: I had accidentally changed this in the last PR, which broke some stuff.  Ooops 😖

handel/src/cli/index.ts
```diff
@@ -321,16 +330,14 @@ export async function checkAction(handelFile: HandelFile, argv: any): Promise<vo
...
 
-    const environmentToDelete = argv.e;
     try {
         await validateLoggedIn();
-        const accountConfig = await config(argv.c); // Load account config to be consumed by the library
+        const accountConfig = await config(options.accountConfig); // Load account config to be consumed by the library
         await validateCredentials(accountConfig);
-        const deleteEnvConfirmed = await confirmDelete(environmentToDelete, argv.y);
-        if (deleteEnvConfirmed) {
+        for (const environmentToDelete of options.environments) {
```
ThatJoeMoore: Turns out that, even though it theoretically supported multiple environments, handel delete didn't.

dsw88: Yes I started out supporting multiple environment deletes intentionally, but then rolled back to only allowing a single one at a time. The reasoning for this was that delete is such a dangerous operation (deleting dbs, buckets, file mounts, etc.) that you should have to do the environment deletes one at a time rather than being able to delete multiples in bulk.

What do you think about that reasoning? I'm inclined to be overly cautious here personally, but maybe I'm just being too paranoid.

ThatJoeMoore: @dsw88 done

dsw88:
It looks like some of the stuff in the cli.validateDeployArgs and cli.validateDeleteArgs could be taken out since they are now being checked higher up. For example, checking that the account config and environments parameters are there could be removed, since you're now checking that in the top-level handel.ts file. 

Some of that logic in those functions still applies though, like checking that the environments are valid in the Handel file.

ThatJoeMoore:
@dsw88 Ahh, yeah, I didn't think of that.  It's been stripped down to the essentials now.

handel/test/lifecycles/delete-test.ts
```diff
@@ -49,8 +49,9 @@ describe('delete lifecycle module', () => {
                 A: new UnPreDeployContext(serviceContext)
             });
             const serviceRegistry = new FakeServiceRegistry();
+            const opts: DeleteOptions = { debug: false, linkExtensions: false, yes: false, environments: ['dev'], accountConfig: '' };
```
dsw88: One more little thing, this module fails to compile for me locally because this line has "environments" still instead of "environment".

dsw88:
I exercised the CLI a little bit and it mostly works great. Here are some things that don't work super great:

First, running without specifying a lifecycle ought to show some top-level help about which lifecycles you can choose
```
handel
```
This produced no output and a status code of 0. It ought to produce some help output about which top-level commands are available.

Second, this command involved a typo:
```
handel deploy -c default-us-west-2 -ee dev
```
Notice the "-ee" instead of "-e". I would expect this to produce some error output of some sort, but it produced this:
```
The following required parameters were omitted: account-config, environments
(node:27905) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): TypeError: command.outputHelp is not a function
(node:27905) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```
The first line looks great, but note the next two lines about unhandled promise rejections. It might be nice if we could figure out what's causing that.

ThatJoeMoore:
Yeah, I probably should have flagged that commit as a WIP.  It's done now, and I've got our winston logger getting configured to match the Caporal one, so we now respect the Caporal-standard --verbose and --quiet flags.  Huzzah!

ThatJoeMoore:
And yeah, I got frustrated enough with Commander (how many things can you have that affect the order of parameters in a callback?  If it's more than 0, you're doing things wrong!), so I searched around for an alternative and found Caporal.  It seems kind of young still, but it's still a step up from raw minimist and has a better API.

dsw88:
I anxiously await the creative title for Handel Extensions Episode III!

PR #411, opened 2018-04-06
Fix problem with parsing CLI tags

ThatJoeMoore:
CLI now uses the appropriate function to parse the tags passed with -t.


dsw88:
Sorry this took a while to merge, I was in a meeting.

PR #413, opened 2018-04-10
Handel Extensions Episode III: Extensions of the Sith

ThatJoeMoore:
# Rise, Lord Extensions

Rise, Lord Extensions

Extensions are here!

This PR touches basically every file in the project, mostly due to moving all of the *Context types to the extensions API.

But extensions are here!

Current Limitations:

- Events don't work for services in extensions
- We don't currently validate the Handel version against the 'engines' in the extension package.json
- We ought to have a better API around outputs/dependency wiring that's a bit more flexible.

BUT THEY'RE HERE!


handel/src/extensions-support/npm/index.ts
```diff
@@ -0,0 +1,76 @@
...
+
+}
+
+export interface LinkedPackage {
+    name: string;
+    path: string;
+}
+
+async function run(tag: string, args: string[], quiet: boolean = false, opts?: SpawnOptions) {
+    const promise = spawn('npm', args, opts);
```
dsw88: Somewhere in this file it would be nice to have a check to make sure npm is available on the path. It's not a guarantee that we'll be able to find it, so it might be nice to make sure it's on the path and show a pretty error if it isn't.

dsw88:
I reviewed this and it looks good, so I'm going to merge it! There are two things I'd like to do soon, in addition to your list of three items already listed in the PR description:

* Make the NPM loader check for NPM on the path prior to trying to use it, and show a pretty error message if it can't find it.
* Add docs around developing extensions

PR #414, opened 2018-04-10
Change Object.values to use Object.keys, which is supported by Node 6

dsw88:


dsw88:
Didn't even think to check whether Lodash offered that...too many choices! I updated to use lodash.values.

PR #424, opened 2018-04-21
extension-support package

dsw88:
This change adds an extension-support package to the Handel project. It is mostly comprised of stuff that used to be in the common/ directory in Handel. 

The purpose of this package is to provide common functionality that should be used by extensions when implementing provisioners. This includes things like the following:

* The names to use for resources and stacks
* Uploading to a common Handel bucket for deployable artifacts.
* Creating security groups and binding to them.
* Deleting stacks and other items.

I don't like the interface exposed by the extension-support package, and I'd like to find a better way. We have to use the "import * as moduleName" syntax in some fashion because you can't really stub functions that are imported with the "import { funcName }" syntax.

For now these are enough changes that I figured we could review and merge them, then make refinements to the module in a subsequent PR. I already have several items that I'd like to improve.


dsw88:
@ThatJoeMoore 174 files changed. Happy weekend! :)

dsw88:
This is not a very glamorous PR, most of the lines changed are just moving stuff around. 

The key thing is that I did take some of the common methods and put them inside the ServiceContext and DeployContext objects, because they really seemed to belong there.

extension-support/test/aws/s3-calls-test.ts
```diff
@@ -0,0 +1,267 @@
...
+
+            const uploadStub = sandbox.stub(awsWrapper.s3, 'upload').resolves({});
+
+            const uploadResponse = await s3Calls.uploadFile('handel-fake-bucket', 'my-key', filePath);
+            expect(uploadResponse).to.deep.equal({});
+            expect(uploadStub.callCount).to.equal(1);
+        });
+    });
+
+    describe('uploadDirectory', () => {
```
ThatJoeMoore: Heads up - Instead of commenting out tests, you can do this:

```javascript
describe.skip('description', () => {
  it('does stuff', () => {})
})
```

[Mocha Docs](https://mochajs.org/#inclusive-tests)

dsw88: Cool, didn't know that! I forgot these were commented out from during the TypeScript migration, so I'm fixing these tests to put them back in.

extension-api/extension-api.ts
```diff
@@ -363,8 +426,18 @@ export class DeployContext implements IDeployContext {
...
+            .reduce((obj: any, name) => {
+                const envName = this.getInjectedEnvVarName(name);
+                obj[envName] = envVars[name];
+                return obj;
+            }, {});
+        Object.assign(this.environmentVariables, formattedVars);
+    }
+
+    public getInjectedEnvVarName(suffix: string) {
+        return `${this.serviceName}_${suffix}`.toUpperCase().replace(/-/g, '_');
```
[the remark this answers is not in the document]
dsw88: Yeah I believe we only allow alphanumeric plus '-' in names in Handel, thanks to the many, many AWS resource naming restrictions across services, so I think this works for me personally rather than pulling in constant-case.

dsw88:
I did end up modifying the library imports, so I pushed up another commit for that. Since it's just modifying the library imports and the review has already been done, I'm going to merge once the tests pass.

PR #432, opened 2018-05-08
Fix broken links

jvisker:


dsw88:
Thanks for fixing these @jvisker!

PR #434, opened 2018-05-14
Change events to use a more well-defined eventOutputs object

dsw88:
This change makes the eventOutputs object in the DeployContext a typed object rather than just a bag of data. 

It also changes the event methods to use a ServiceEventType object instead of the service provisioner type. This allows an extension to be written that "provides" a particular service event type and still work with other provisioners that accept that kind of service event.

This change also introduces a DeployOutputType enum so that we don't just use strings for it in the TS code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ coverage/
 dist/
 lerna-debug.log
 .vscode/
+.handel-extensions/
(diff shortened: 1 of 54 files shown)


dsw88:
@ThatJoeMoore I sort of showed this to you in-person a few days ago, do you want to review this in detail, or should I just merge since I already talked with you in-person about this change?

PR #435, opened 2018-05-16
added support for windows npm

yoshutch:
When trying to create handel extensions on my windows machine we ran into an error where my machine was trying to run npm the linux way and not the windows way. This should check if you're running on windows, if so then it will use the `npm.cmd` instead of `npm`


dsw88:
Thanks @yoshutch!

PR #439, opened 2018-06-07
Update Gulp to 4.0.0

ThatJoeMoore:
Gulp 3.9.1 depends on graceful-fs, which depends on 'natives', which is
an evil package and subject to breakage at any time by the Node team,
which they did in the latest version of Node 10.  Gulp 4 has a newer
verison of graceful-fs, which doesn't work as much evil in the land.

I had to update the gulpfiles a bit in order to make this work.

I'm wondering what utility Gulp is really giving us? I'm pretty sure
everything it's doing could be accomplished using npm scripts that call
into tsc and rimraf. We should probably explore removing our dependency
on gulp, since it's a pretty big dependency.


dsw88:
I merged this, but I agree that we should remove Gulp soon. I originally used it following another project that was using it, but it definitely isn't providing any value and needs to be removed.

PR #449, opened 2018-07-06
Add a nicer message when Route53 zones are missing

ThatJoeMoore:
This also unifies the message across all places where we're searching
for a zone ID.

It may make more sense to create another method that wraps the
getBestMatchingHostedZone function, which throws if there isn't one.  I
don't necessarily want to do that inside of getBestMatchingHostedZone,
because I can see cases where we'd want to use the falsy-returning
version.  Perhaps a method named 'requireBestMatchingHostedZone'?


dsw88:
Thanks for fixing this! Haha this continues the work of removing most of the `!` usages from the codebase. I added multiple of them during the migration to shorten the work I had to do, but I find that in almost all cases the compiler did indeed know better than me that something would be undefined at some point. :)

PR #452, opened 2018-07-09
Implement Automatic Lambda Warming

ThatJoeMoore:
Adds the ability to schedule automatic lambda warmup calls to API Gateway resources. Optionally will make the call simulate an HTTP API Gateway Proxy request.

### DIFF
--- a/extension-support/src/util/handlebars-utils.ts
+++ b/extension-support/src/util/handlebars-utils.ts
@@ -33,7 +33,7 @@ export function compileTemplate(filename: string, variables: any): Promise<strin
     return new Promise((resolve, reject) => {
         fs.readFile(filename, 'utf-8', (error, source) => {
             // Register any helpers we need
-            const template = handlebars.compile(source);
+            const template = handlebars.compile(source, {noEscape: true});
             const output = template(variables);
             resolve(output);
         });
--- a/handel/src/services/apigateway/proxy/apigateway-proxy-template.yml
+++ b/handel/src/services/apigateway/proxy/apigateway-proxy-template.yml
@@ -152,6 +152,35 @@ Resources:
           DNSName: !GetAtt CustomDomain{{logicalId name}}.DistributionDomainName
   {{/each}}
 
+  {{#if warmup}}
+  LambdaWarmupPermission:
+    Type: "AWS::Lambda::Permission"
+    Properties:
+      Action: 'lambda:InvokeFunction'
+      FunctionName: !Ref LambdaFunction
+      Principal: events.amazonaws.com
+
+  LambdaWarmupEvent:
+    Type: "AWS::Events::Rule"
+    Properties:
+      Description: "Handel-created warmup rule for {{apiName}}"
+      Name: "{{apiName}}-warmup"
+      ScheduleExpression: "{{warmup.schedule}}"
+      State: ENABLED
+      Targets:
+      {{#unless warmup.httpPaths}}
+        - Id: {{apiName}}
+          Arn: !GetAtt LambdaFunction.Arn
+      {{else}}
+      {{#each warmup.httpPaths}}
+        - Id: {{logicalId ../apiName}}{{logicalId path}}
+          Arn: !GetAtt LambdaFunction.Arn
+          Input:
+            Fn::Sub: {{eventBody}}
+      {{/each}}
+      {{/unless}}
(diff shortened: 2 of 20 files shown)


extension-support/src/util/handlebars-utils.ts
```diff
@@ -33,7 +33,7 @@ export function compileTemplate(filename: string, variables: any): Promise<strin
     return new Promise((resolve, reject) => {
         fs.readFile(filename, 'utf-8', (error, source) => {
             // Register any helpers we need
-            const template = handlebars.compile(source);
+            const template = handlebars.compile(source, {noEscape: true});
```
dsw88: Wow I didn't know this parameter was a thing! I've just been using the triple-bracket method in places where the escaping could be a problem: `{{{value}}}`. I much prefer this though so we can just use the double-bracket syntax everywhere.

handel/src/services/apigateway/proxy/apigateway-proxy-template.yml
```diff
@@ -152,6 +152,35 @@ Resources:
           DNSName: !GetAtt CustomDomain{{logicalId name}}.DistributionDomainName
   {{/each}}
 
+  {{#if warmup}}
+  LambdaWarmupPermission:
+    Type: "AWS::Lambda::Permission"
+    Properties:
+      Action: 'lambda:InvokeFunction'
+      FunctionName: !Ref LambdaFunction
+      Principal: events.amazonaws.com
```
dsw88: Can this have a resource placed on it? I know some don't, but I thought I remembered that these can be constrained to the individual event defined in the template.

handel/src/services/apigateway/proxy/proxy-passthrough-deploy-type.ts
```diff
@@ -77,6 +79,22 @@ async function getCompiledApiGatewayTemplate(stackName: string, ownServiceContex
         handlebarsParams.customDomains = await apigatewayCommon.getCustomDomainHandlebarsParams(ownServiceContext, serviceParams.custom_domains);
     }
 
+    const warmup: WarmupConfig = getParam(serviceParams, 'warmup', 'warmup', undefined);
```
dsw88: Just as an FYI, I'm hopefully going to be remove the old deprecated configuration syntax for proxy APIs here in the next few days (almost everyone has moved to the new configuration style), so we'll be able to get rid of that `getParam` method to make things a tiny bit clearer.

handel/src/services/apigateway/swagger/apigateway-swagger-template.yml
```diff
@@ -92,6 +92,36 @@ Resources:
...
       {{/each}}
       {{/if}}
+
+  {{#if warmup}}
+  LambdaWarmupPermission{{@index}}:
+    Type: "AWS::Lambda::Permission"
+    Properties:
+      Action: 'lambda:InvokeFunction'
+      FunctionName: !Ref LambdaFunction{{@index}}
+      Principal: events.amazonaws.com
```
dsw88: Like in the other template, it would be great if we could put a resource constraint here if possible.

handel/src/services/apigateway/swagger/swagger-deploy-type.ts
```diff
@@ -253,8 +288,23 @@ export async function deploy(stackName: string, ownServiceContext: ServiceContex
     lambdasToCreate = await uploadDeployableArtifactsToS3(ownServiceContext, lambdasToCreate, serviceName, enrichedSwagger);
     const swaggerS3ArtifactInfo = await uploadSwaggerToS3(ownServiceContext, enrichedSwagger);
     const compiledTemplate = await getCompiledApiGatewayTemplate(stackName, ownServiceContext, ownPreDeployContext, dependenciesDeployContexts, lambdasToCreate, swaggerS3ArtifactInfo, stackTags);
+
+    await require('fs-extra').writeFile('./apigw.yml', compiledTemplate);
```
dsw88: What is the purpose of this line?

ThatJoeMoore: The purpose of that line is that I was supposed to remove it before committing 🤦‍♂️

ThatJoeMoore:
Everything should be ready to review.

PR #483, opened 2018-07-19
Remove Gulp from project

dsw88:
This change removes Gulp in favor of simple scripts that accomplish the same task.


dsw88:
@ThatJoeMoore I've updated to use the CLI tools you recommended, let me know how it looks now.

dsw88:
@ThatJoeMoore I heard you're out for a while. I think that I've addressed your previous comments, and the build process is still working properly, so I'm going to merge for now. 

Definitely still take a look at the final changes to see if there are things that should be modified, and I will address them post-merge in another commit.

PR #505, opened 2018-08-14
Made a schema and appropriate tests for route53zone

### LINKED ISSUE #472: Route53 Zone - Add JSON Schema to Check phase

mxnguyen350:
Resolves #472

### DIFF
--- a/examples/route53zone/handel.yml
+++ b/examples/route53zone/handel.yml
@@ -1,0 +1,9 @@
+version: 1
+name: route53zone-test
+environments:
+  dev:
+    zone:
+      type: route53zone
+      name: mydomain.example.com
+      private: true
+      tags: mytagvalue
--- a/examples/route53zone/handel.yml
+++ b/examples/route53zone/handel.yml
@@ -6,4 +6,5 @@ environments:
       type: route53zone
       name: mydomain.example.com
       private: true
-      tags: mytagvalue
+      tags:
+        mytagename: mytagvalue
--- a/handel/src/services/route53zone/index.ts
+++ b/handel/src/services/route53zone/index.ts
@@ -78,6 +78,10 @@ export function check(serviceContext: ServiceContext<Route53ZoneServiceConfig>, 
     const errors: string[] = checkPhase.checkJsonSchema(`${__dirname}/params-schema.json`, serviceContext);
 
     const params = serviceContext.params;
+    
+    if (!route53.isValidHostname(params.name)) {
+    errors.push(`'name' parameter must be a valid hostname`);
+}
 
     return errors.map(error => `${SERVICE_NAME} - ${error}`);
 }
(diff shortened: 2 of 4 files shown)


examples/route53zone/handel.yml
```diff
@@ -0,0 +1,9 @@
+version: 1
+name: route53zone-test
+environments:
+  dev:
+    zone:
+      type: route53zone
+      name: mydomain.example.com
+      private: true
+      tags: mytagvalue
```
dsw88: I think this ought to fail, it should be something like this:
```
tags:
  mytagname: mytagvalue
```

handel/src/services/route53zone/index.ts
```diff
@@ -75,21 +75,21 @@ function getCompiledRoute53Template(ownServiceContext: ServiceContext<Route53Zon
  */
 
 export function check(serviceContext: ServiceContext<Route53ZoneServiceConfig>, dependenciesServiceContexts: Array<ServiceContext<ServiceConfig>>): string[] {
-    const errors = [];
+    const errors: string[] = checkPhase.checkJsonSchema(`${__dirname}/params-schema.json`, serviceContext);
 
     const params = serviceContext.params;
 
     if (!params.name) {
```
dsw88: This first if branch can be removed because it's now checked in the JSON schema.

handel/src/services/route53zone/index.ts
```diff
@@ -75,21 +75,21 @@ function getCompiledRoute53Template(ownServiceContext: ServiceContext<Route53Zon
...
-        errors.push(`${SERVICE_NAME} - 'name' parameter must be specified`);
+        errors.push(`'name' parameter must be specified`);
     } else if (!route53.isValidHostname(params.name)) {
-        errors.push(`${SERVICE_NAME} - 'name' parameter must be a valid hostname`);
+        errors.push(`'name' parameter must be a valid hostname`);
     }
 
     if (params.private && typeof params.private !== 'boolean') {
```
dsw88: This if statement should also be able to be removed because it's checked in the JSON schema now.

handel/src/services/route53zone/index.ts
```diff
@@ -75,21 +75,11 @@ function getCompiledRoute53Template(ownServiceContext: ServiceContext<Route53Zon
...
 
 export function check(serviceContext: ServiceContext<Route53ZoneServiceConfig>, dependenciesServiceContexts: Array<ServiceContext<ServiceConfig>>): string[] {
-    const errors = [];
+    const errors: string[] = checkPhase.checkJsonSchema(`${__dirname}/params-schema.json`, serviceContext);
 
     const params = serviceContext.params;
 
-    if (!params.name) {
-        errors.push(`${SERVICE_NAME} - 'name' parameter must be specified`);
-    } else if (!route53.isValidHostname(params.name)) {
```
dsw88: Sorry I meant just the first "if" part needed to go. The check inside the "else if" needs to stay. Sorry for not being clearer on that!

PR #506, opened 2018-08-14
Added schema and needed tests to iot

mxnguyen350:


handel/src/services/iot/index.ts
```diff
@@ -86,21 +86,21 @@ async function deleteTopicRule(ruleName: string) {
  */
 
 export function check(serviceContext: ServiceContext<IotServiceConfig>, dependenciesServiceContexts: Array<ServiceContext<ServiceConfig>>) {
-    const errors = [];
-
+    const errors: string[] = checkPhase.checkJsonSchema(`${__dirname}/params-schema.json`, serviceContext);
+    
     const serviceParams = serviceContext.params;
     if (serviceParams.event_consumers) {
```
mxnguyen350: I wasn't sure if these if statements needed to be removed, so I left them in

dsw88: Oh yes I need to talk to you about these event_consumers. They are one of two special parameter types that can be applied to multiple services, so in this case we'll need to include them in the JSON schema.

handel/src/services/iot/params-schema.json
```diff
@@ -0,0 +1,68 @@
...
+                        "type": "string",
+                        "errorMessage": "The 'sql' parameter is required in each config in the 'event_consumers' section"
+                    }
+                },
+                "required": [
+                    "service_name",
+                    "sql"
+                ],
+                "errorMessage": {
+                    "required": {
```
dsw88: You'll want to add an error message for "additionalProperties" here too in addition to the one you already have for "required".

Issue #466, opened 2018-07-16
KMS - Add JSON Schema to Check phase


dsw88:

dsw88:
Resolved by https://github.com/byu-oit/handel/pull/507

PR #507, opened 2018-08-16
Updated error message in ses and created schema for ksm

mxnguyen350:
Resloves #466

### DIFF
--- a/examples/ksm/handel.yml
+++ b/examples/ksm/handel.yml
@@ -1,0 +1,10 @@
+version: 1
+name: ksm-test
+environments:
+  dev:
+    key:
+      type: ksm
+      alais: examplealias3
+      auto_rotate: true
+      tags:
+        mytagname: mytagvalue
--- a/handel/src/services/kms/index.ts
+++ b/handel/src/services/kms/index.ts
@@ -90,7 +90,7 @@ function getDefaultAlias(serviceContext: ServiceContext<KmsServiceConfig>): stri
  */
 
 export function check(serviceContext: ServiceContext<KmsServiceConfig>, dependenciesServiceContexts: Array<ServiceContext<ServiceConfig>>): string[] {
-    const errors = [];
+    const errors: string[] = checkPhase.checkJsonSchema(`${__dirname}/params-schema.json`, serviceContext);
 
     const params = serviceContext.params;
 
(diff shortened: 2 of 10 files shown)


handel/src/services/kms/index.ts
```diff
@@ -90,7 +90,7 @@ function getDefaultAlias(serviceContext: ServiceContext<KmsServiceConfig>): stri
  */
 
 export function check(serviceContext: ServiceContext<KmsServiceConfig>, dependenciesServiceContexts: Array<ServiceContext<ServiceConfig>>): string[] {
-    const errors = [];
+    const errors: string[] = checkPhase.checkJsonSchema(`${__dirname}/params-schema.json`, serviceContext);
```
dsw88: You're missing the import for the `checkPhase` module, so the compile step of the build is failing. Make sure to run `npm run build` when you make changes prior to pushing up the branch; that will help you catch these kinds of compile errors.

handel/src/services/kms/params-schema.json
```diff
@@ -0,0 +1,44 @@
+{
+    "$schema": "http://json-schema.org/schema#",
+    "type": "object",
+    "properties": {
+        "type": {
+            "type": "string",
+            "description": "The service type (should always be 'ses')",
```
dsw88: Looks like the 'ses' on this line should probably be changed to 'kms'.

handel/src/services/kms/params-schema.json
```diff
@@ -0,0 +1,44 @@
...
+            "pattern": "ksm",
```
dsw88: Typo here and on the next line, it should be 'kms' instead of 'ksm'

handel/src/services/kms/params-schema.json
```diff
@@ -0,0 +1,44 @@
...
+            "errorMessage": "The 'type' field must equal 'ksm'"
+        },
+        "alias": {
+            "type": "string",
+            "errorMessage": "The 'alais' parameter must be a string",
```
dsw88: Typo in the word 'alias'.

examples/ksm/handel.yml
```diff
@@ -0,0 +1,10 @@
+version: 1
+name: ksm-test
+environments:
+  dev:
+    key:
+      type: ksm
```
dsw88: The type here should be 'kms' instead of

PR #509, opened 2018-08-17
Sqs schema

### LINKED ISSUE #477: SQS - Add JSON Schema to Check phase

mxnguyen350:
Resolves #477

### DIFF
--- a/handel/src/services/sqs/params-schema.json
+++ b/handel/src/services/sqs/params-schema.json
@@ -1,0 +1,130 @@
+{
+    "$schema": "http://json-schema.org/schema#",
+    "type": "object",
+    "properties": {
+        "type": {
+            "type": "string",
+            "description": "The service type (should always be 'sqs')",
+            "pattern": "sqs",
+            "errorMessage": "The 'type' parameter must equal 'sqs'"
+        },
+        "query_type": {
+            "type": "string",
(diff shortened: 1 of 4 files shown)


handel/src/services/sqs/params-schema.json
```diff
@@ -0,0 +1,130 @@
+{
+    "$schema": "http://json-schema.org/schema#",
+    "type": "object",
+    "properties": {
+        "type": {
+            "type": "string",
+            "description": "The service type (should always be 'sqs')",
+            "pattern": "sqs",
```
dsw88: You'll want to put the '^' and '$' characters on the two ends of this pattern so that it will only accept the string 'sqs'

handel/src/services/sqs/params-schema.json
```diff
@@ -0,0 +1,130 @@
...
+            "properties": {
+                "max_receive_count":{
+                    "type": "number",
+                    "errorMessage": "The 'max_receive_count' field must be a number"
+                },
+                "delay_seconds": {
+                    "type": "number",
+                    "errorMessage": "The 'delay_seconds' field must be a number"
+                },
+                "max_message_size": {
```
dsw88: I love that you put minimum and maximums on the parameters at the top-level, it would be great if we could add those for the number types in the 'dead_letter_queue' section as well.

PR #511, opened 2018-08-21
Made a JSON schema for s3

### LINKED ISSUE #473: S3 - Add JSON Schema to Check phase

mxnguyen350:
Resolves #473


handel/src/services/s3/index.ts
```diff
@@ -144,20 +144,15 @@ function getS3EventFilters(filterList: S3ServiceEventFilterList | undefined): AW
  */
 
 export function check(serviceContext: ServiceContext<S3ServiceConfig>, dependenciesServiceContexts: Array<ServiceContext<ServiceConfig>>): string[] {
-    const errors = [];
+    const errors: string[] = checkPhase.checkJsonSchema(`${__dirname}/params-schema.json`, serviceContext);
+
     const validAcls = ['AuthenticatedRead', 'AwsExecRead', 'BucketOwnerRead', 'BucketOwnerFullControl', 'LogDeliveryWrite', 'Private', 'PublicRead'];
```
dsw88: This line and the one after it can be deleted because they're not referenced anywhere anymore after deleting those two if statements below.

PR #519, opened 2018-08-23
added JSON schema to dynamodb

### LINKED ISSUE #461: DynamoDB - Add JSON Schema to Check phase

mxnguyen350:
Resolves #461

### DIFF
--- a/handel/src/services/dynamodb/index.ts
+++ b/handel/src/services/dynamodb/index.ts
@@ -37,6 +37,7 @@ import {
 import * as winston from 'winston';
 import * as autoscaling from './autoscaling';
 import {DynamoDBConfig} from './config-types';
+import { CloudFormation } from 'aws-sdk';
 
 const KEY_TYPE_TO_ATTRIBUTE_TYPE: any = {
     String: 'S',
@@ -80,7 +81,7 @@ function buildTableARN(tableName: string, accountConfig: AccountConfig) {
 
 function getDeployContext(serviceContext: ServiceContext<DynamoDBConfig>, cfStack: AWS.CloudFormation.Stack): DeployContext {
     const deployContext = new DeployContext(serviceContext);
-    const tableName = awsCalls.cloudFormation.getOutput('TableName', cfStack);
+    const tableName = getTableNameFrom(cfStack);
     if(!tableName) {
         throw new Error('Expected to receive TableName back from DynamoDB service');
     }
@@ -108,6 +109,10 @@ function getDeployContext(serviceContext: ServiceContext<DynamoDBConfig>, cfStac
     });
 
     return deployContext;
+}
+
+function getTableNameFrom(stack: CloudFormation.Stack): string {
+    return awsCalls.cloudFormation.getOutput('TableName', stack)!;
 }
 
 function addDefinedAttribute(definedAttrs: any[], attrName: string, attrType: string) {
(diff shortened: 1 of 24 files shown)


handel/src/services/dynamodb/index.ts
```diff
@@ -239,81 +246,16 @@ async function getCompiledDynamoTemplate(stackName: string, ownServiceContext: S
...
-        if (!params.partition_key.name) {
-            errors.push(`${SERVICE_NAME} - The 'name' field in the 'partition_key' section is required`);
-        }
-        if (!params.partition_key.type) {
-            errors.push(`${SERVICE_NAME} - The 'type' field in the 'partition_key' section is required`);
-        }
-    }
-
-    // Check throughput
-    errors.push(...autoscaling.checkProvisionedThroughput(params.provisioned_throughput, `${SERVICE_NAME} - `));
```
dsw88: It looks like you're removing this line but I don't see a corresponding removal of the method itself in the autoscaling module. We should probably look at this method and decide whether to keep it. If we keep it we can put this line back in or if we don't need it then we can delete the method.

handel/src/services/dynamodb/params-schema.json
```diff
@@ -0,0 +1,361 @@
...
+                    "errorMessage": "The 'name' parameter must be a string"
+                },
+                "type": {
+                    "type": "string",
+                    "description": "The 'type' parameter may only be 'String' or 'Number'",
+                    "pattern": "^String|Number$",
+                    "errorMessage": "The 'type' parameter must be 'String' or 'Number'"
+                }
+            },
+            "additionalProperties": false,
```
dsw88: We'll want to add the "required" section in here for the "name" and "type" parameters inside "sort_key".

handel/src/services/dynamodb/params-schema.json
```diff
@@ -0,0 +1,361 @@
...
+                    "errorMessage": "The 'read_target_utilization' parameter must be a percentage"
+                },
+                "write_target_utilization": {
+                    "type": "number",
+                    "minimum": 0,
+                    "maximum": 100,
+                    "errorMessage": "The 'write_target_utilization' parameter must be a percentage"
+                }
+            },
+            "additionalProperties": false,
```
dsw88: We'll want to have the "required" section here for "read_capacity_units" and "write_capacity_units"

handel/src/services/dynamodb/params-schema.json
```diff
@@ -0,0 +1,361 @@
...
+                                "errorMessage": "The 'name' parameter must be a string"
+                            },
+                            "type": {
+                                "type": "string",
+                                "description": "The 'type' parameter may only be 'String' or 'Number'",
+                                "pattern": "^String|Number$",
+                                "errorMessage": "The 'type' parameter must be 'String' or 'Number'"
+                            }
+                        },
+                        "additionalProperties": false,
```
dsw88: This section needs "name" and "type" to be required.

handel/src/services/dynamodb/params-schema.json
```diff
@@ -0,0 +1,361 @@
...
+                        "errorMessage": {
+                            "additionalProperties": "Invalid/Unknown property specified in the 'provisioned_throughput' section of 'global_indexes'"
+                        }
+                    }
+                },
+                "required": [
+                    "name",
+                    "partition_key",
+                    "sort_key",
```
dsw88: I don't think sort_key should be required in the global_indexes section.

handel/test/services/dynamodb/dynamodb-test.ts
```diff
@@ -220,6 +220,7 @@ describe('dynamodb deployer', () => {
                 };
 
                 const errors = dynamodb.check(serviceContext, []);
+                console.log(errors);
```
dsw88: This looks like debug code that should be removed.

PR #525, opened 2018-08-28
made key_types case insensitive

### LINKED ISSUE #277: DynamoDB - Key types should be case-insensitive
"String" is currently required, but "string" should really work properly too. Let's support case-insensitive versions of the key types.

mxnguyen350:
resolves #277

### DIFF
--- a/handel/src/services/dynamodb/params-schema.json
+++ b/handel/src/services/dynamodb/params-schema.json
@@ -25,7 +25,7 @@
                 "type": {
                     "type": "string",
                     "description": "The 'type' parameter may only be 'String' or 'Number'",
-                    "pattern": "^String|Number$",
+                    "pattern": "^[sS][tT][rR][iI][nN][gG]|[nN][uU][mM][bB][eE][rR]$",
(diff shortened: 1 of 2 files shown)


handel/src/services/dynamodb/params-schema.json
```diff
@@ -25,7 +25,7 @@
                 "type": {
                     "type": "string",
                     "description": "The 'type' parameter may only be 'String' or 'Number'",
-                    "pattern": "^String|Number$",
+                    "pattern": "^[sS][tT][rR][iI][nN][gG]|[nN][uU][mM][bB][eE][rR]$",
```
dsw88: I changed my mind, I think we only want to allow lowercase (i.e. 'string') or capital-case (i.e. 'String'). It's really supposed to be 'String' according to DynamoDB but the lowercase variant is just a helper to people who forget to put the capital letter in first.

We may also need to look at modifying some code to make sure this will actually work. Dynamo wants 'String' and 'Number', so we'll have to transform 'string' and 'number' prior to passing them into the CloudFormation template.

PR #527, opened 2018-08-29
fixed error with regex to allow the proper number of characters

mxnguyen350:
Resolves issue mentioned in the slack channel

### DIFF
--- a/handel/src/services/dynamodb/params-schema.json
+++ b/handel/src/services/dynamodb/params-schema.json
@@ -12,7 +12,7 @@
             "type": "string",
             "minLength": 3,
             "maxLength": 255,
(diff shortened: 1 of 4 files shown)


handel/src/services/dynamodb/params-schema.json
```diff
@@ -12,7 +12,7 @@
             "type": "string",
             "minLength": 3,
             "maxLength": 255,
-            "pattern": "^[a-zA-Z0-9_\\-.]$",
+            "pattern": "^[a-zA-Z0-9_\\-.]*$",
```
dsw88: Haha yep that would do it, sorry I didn't catch this in the last review!

PR #534, opened 2018-08-31
added a default region to environment variables

### LINKED ISSUE #412: Set environment variable for AWS region
I was deploying a Fargate application and ran into the problem that the AWS region was defaulted to 'us-west-2'. It would be convenient to have a Handel environment variable that provides the AWS region that you are in. I know that you can make the following request to discover the current AWS region in a deployed process:
`curl --silent http://169.254.169.254/latest/meta-data/hostname | cut -d\. -f2`

But, it would still be very nice for the application to be able to reference the AWS region as a Handel environment variable.

Something like: HANDEL_AWS_REGION or HANDEL_CURRENT_AWS_REGION

mxnguyen350:
Resolves #412

### DIFF
--- a/extension-support/test/common/deploy-phase-test.ts
+++ b/extension-support/test/common/deploy-phase-test.ts
@@ -229,6 +229,8 @@ describe('Deploy phase common module', () => {
             expect(envVars.HANDEL_ENVIRONMENT_NAME).to.equal('FakeEnv');
             expect(envVars.HANDEL_SERVICE_NAME).to.equal('FakeService');
             expect(envVars.HANDEL_PARAMETER_STORE_PREFIX).to.equal('FakeApp.FakeEnv');
+            expect(envVars.TEST_USER_VAR).to.equal('test_value');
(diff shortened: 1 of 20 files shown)


extension-support/test/common/deploy-phase-test.ts
```diff
@@ -211,4 +211,24 @@ describe('Deploy phase common module', () => {
...
+            const dependencyDeployContext = new DeployContext(dependencyServiceContext);
+            dependencyDeployContext.addEnvironmentVariables({
+                TEST_DEPENDENCY_VAR: 'test_value2'
+            });
+            const dependenciesDeployContexts = [dependencyDeployContext];
+            const userProvidedEnvVars = {
+                TEST_USER_VAR: 'test_value'
+            };
+            const envVars = deployPhase.getEnvVarsForDeployedService(serviceContext, dependenciesDeployContexts, userProvidedEnvVars);
+            expect(envVars.HANDEL_REGION_NAME).to.equal('us-west-2');
```
dsw88: We should probably add some more expectations to verify that TEST_USER_VAR and TEST_DEPENDENCY_VAR were properly added to the 'envVars' variable.

PR #537, opened 2018-09-05
added health check to ECS and fargate

### LINKED ISSUE #361: ECS - allow health check grace periods
It would be nice to allow the ECS and probably fargate services to specify the health check grace period. Some apps take a while to spin up and would be nice to wait before checking the health checks.
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-service.html#cfn-ecs-service-healthcheckgraceperiodseconds

mxnguyen350:
Resolves #361

### DIFF
--- a/handel/src/services/ecs-fargate/config-types.ts
+++ b/handel/src/services/ecs-fargate/config-types.ts
@@ -14,6 +14,7 @@
  * limitations under the License.
  *
  */
+import { integer } from 'aws-sdk/clients/lightsail';
 import { ServiceConfig, Tags } from 'handel-extension-api';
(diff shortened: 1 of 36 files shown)


handel/src/services/ecs-fargate/config-types.ts
```diff
@@ -14,6 +14,7 @@
  * limitations under the License.
  *
  */
+import { integer } from 'aws-sdk/clients/lightsail';
```
dsw88: This looks like an accidental import by your editor that can be removed.

PR #551, opened 2018-09-14
Uses Pricing API to get instance type size

### LINKED ISSUE #388: ECS - Use AWS Pricing API instead of hard-coded instance attributes
When I originally implemented the ECS service, I couldn't find a way to get the memory and CPU allocations for each instance type from the AWS APIs. As a result I ended up hard-coding these values in the ECS service. This has the downside of needing to manually update these attributes for new instances before you can use them.

It turns out the AWS pricing API offers this information: https://aws.amazon.com/blogs/aws/new-aws-price-list-api/. 

Here is an example CLI call to get information about the t2.medium:
```
aws pricing get-products --region us-east-1 --service-code AmazonEC2 --filters Type=TERM_MATCH,Field=instanceType,Value=t2.medium "Type=TERM_MATCH,Field=location,Value=US West (Oregon)" --max-items 1
```

The Node SDK supports this API. One of the strange things about this API is that you can only hit it in the us-east-1 or ap-south-1 regions, but you can use either one to get pricing information for all the other regions. When we implement this, we'll need to be careful not to globally set the region to us-east-1, but to only set the pricing API calls to that region and leave the configured region for all other calls.

mxnguyen350:
Resolves #388

### DIFF
--- a/handel/src/aws/pricing-calls.ts
+++ b/handel/src/aws/pricing-calls.ts
@@ -17,14 +17,42 @@
 
 import awsWrapper from './aws-wrapper';
 
-function convertMemoryStringToMemoryUnits(inputString: string) {
+function convertMemoryStringToMemoryUnits(inputString: string): number {
     // Converts the given GiB into the amazon units
-    const gibiBitesOfMemory = inputString.split('GiB', 1);
-    const amazonMemoryUnits = Number(gibiBitesOfMemory) * 1000;
+    const gibiBitesOfMemory = inputString.split('GiB', 1)[0];
+    const amazonMemoryUnits = Number(gibiBitesOfMemory.trim()) * 1000;
     return amazonMemoryUnits;
 }
 
-export async function getMemoryForInstance(instanceType: string): Promise<number> {
+interface AWSRegionMap {
+    [key: string]: string;
+}
+
+const AWS_REGION_MAP: AWSRegionMap = {
+    'us-west-2': 'US West (Oregon)',
+    'us-west-1': 'US West (N. California)',
+    'us-east-2': 'US East (Ohio)',
+    'es-east-1': 'US East (N. Virginia)',
+    'ap-south-1': 'Asia Pacific (Mumbai)',
+    'ap-northeast-2': 'Asia Pacific (Seoul)',
+    'ap-northeast-1': 'Asia Pacific (Tokyo)',
+    'ap-southeast-1': 'Asia Pacific (Singapore)',
+    'ap-southeast-2': 'Asia Pacific (Sydney)',
+    'ca-central-1': 'Canada (Central)',
+    'cn-north-1': 'China (Beijing)',
+    'eu-central-1': 'EU (Frankfurt)',
+    'eu-west-1': 'EU (Ireland)',
+    'eu-west-2': 'EU (London)',
+    'eu-west-3': 'EU (Paris)',
+    'sa-east-1': 'South America (São Paulo)',
+    'us-gov-west-1': 'AWS GovCloud (US)'
+};
+
+export async function getMemoryForInstance(instanceType: string, region: string): Promise<number> {
+    const regionPricingName = AWS_REGION_MAP[region];
+    if (!regionPricingName) {
+        throw new Error(`Invalid/Unknown region name specified: ${region}`);
+    }
     const getProductsParams = {
         ServiceCode: 'AmazonEC2',
         Filters: [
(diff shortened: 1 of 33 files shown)


handel/src/aws/pricing-calls.ts
```diff
@@ -0,0 +1,57 @@
...
+        Filters: [
+            {
+                Type: 'TERM_MATCH',
+                Field: 'instanceType',
+                Value: instanceType
+            },
+            {
+                Type: 'TERM_MATCH',
+                Field: 'location',
+                Value: 'US West (Oregon)'
```
dsw88: This needs to be changed to support any region, since Handel can deploy to any region. We'll probably need a mapping from region names like 'us-west-2' to the value they want like 'US West (Oregon)'.

Better yet, we should check the API for GetProducts agains to see if there is a filter that will let us give the actual region name like 'us-west-2'.

handel/src/aws/pricing-calls.ts
```diff
@@ -0,0 +1,57 @@
...
+            }
+        ],
+        MaxResults: 1
+    };
+    try {
+        const response = await awsWrapper.pricing.getProducts(getProductsParams);
+        if (!response.PriceList || !response.PriceList[0]) {
+            throw new Error(`Price list does not exits for instance type specified: ${instanceType}`);
```
dsw88: Spelling error here: "exits" should be "exist"

handel/src/aws/pricing-calls.ts
```diff
@@ -0,0 +1,57 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+import awsWrapper from './aws-wrapper';
+
+function convertMemoryStringToMemoryUnits(inputString: string) {
+    // Converts the given GiB into the amazon units
+    const gibiBitesOfMemory = inputString.split('GiB', 1);
```
dsw88: It's probably good practice to call .trim() after doing the split so that you just pass the number in and no whitespace. It probably doesn't affect functionality here, but it could be used in future modifications where that would be important.

handel/src/aws/pricing-calls.ts
```diff
@@ -0,0 +1,57 @@
...
+        MaxResults: 1
+    };
+    try {
+        const response = await awsWrapper.pricing.getProducts(getProductsParams);
+        if (!response.PriceList || !response.PriceList[0]) {
+            throw new Error(`Price list does not exits for instance type specified: ${instanceType}`);
+        }
+        const priceList = response.PriceList[0] as any;
+        const memory: string = priceList.product.attributes.memory; // Something like '4 GiB'
+        const memoryUnits: number = convertMemoryStringToMemoryUnits(memory); // Convert '4 GiB' to 4000
```
dsw88: It's probably better to make the convertMemoryStringToMemoryUnits() function explicitly have a return type of `number`, then you can remove the `number` type definition on this line when you declare the `memoryUnits` variable.

I think it's better to declare the type on the function because then you're explicitly saying the function must return a number, so it can help let you know whether or not you're honoring that contract in the function body.

handel/src/aws/pricing-calls.ts
```diff
@@ -0,0 +1,57 @@
...
+        return memoryUnits;
+    } catch (e) {
+        // Maybe we want to log the error?
+        throw new Error(`Could not run code to find price list for the instance type specified: ${instanceType}`);
```
dsw88: Maybe tweak this error a little bit to be a tad clearer: 
```
Could not obtain memory information for the instance type '${instanceType}'
```

handel/src/services/ecs/cluster-auto-scaling.ts
```diff
@@ -150,11 +83,8 @@ export async function createDrainingLambdaIfNotExists(accountConfig: AccountConf
  *
  * This function is used for both 'min' and 'max' auto-scaling group calculations.
  */
-export function getInstanceCountForCluster(instanceType: string, autoScaling: HandlebarsEcsTemplateAutoScaling, containerConfigs: HandlebarsEcsTemplateContainer[], calculationType: string, serviceName: string): number {
-    const instanceMemory = EC2_INSTANCE_MEMORY_MAP[instanceType];
-    if (!instanceMemory) {
-        throw new Error(`${serviceName} - Unhandled instance type specified: ${instanceType}`);
-    }
+export async function getInstanceCountForCluster(instanceType: string, autoScaling: HandlebarsEcsTemplateAutoScaling, containerConfigs: HandlebarsEcsTemplateContainer[], calculationType: string, serviceName: string): Promise<number> {
+    const instanceMemory = await getMemoryForInstance(instanceType);
```
dsw88: This call takes a little bit of time, and we call this `getInstanceCountForCluster` twice for the same instance type. We should probably call this getMemoryForInstance() outside this function and pass the value it returns into both invocations of this function. That will result in a small speed-up at run-time.

PR #553, opened 2018-09-14
Modify UnBind Phase Contract

dsw88:
I had hoped it was sufficient to just pass the ServiceContext to UnBind, and it has been for a while. However, as I write an extension, I find myself needing to UnBind just one service combination at a time, rather than all at once as we have previously been doing. In order to do this, it requires information from the PreDeployContexts of both services.

I can't think of any way to get this information except by running the PreDeploy phase first prior to running the existing delete phases. While this feels a bit odd, it very nicely provides the information from PreDeploy to the delete phases that need it.

This is a breaking extension contract change at compile-time (not at run-time because we only add more parameters onto the end of the UnBind method). Luckily there are few enough extensions out there currently that we can go modify them without having to call this a major version change.

@ThatJoeMoore Let me know if you can think of a better way of doing this.

### DIFF
--- a/docs/extensions/writing-extensions.rst
+++ b/docs/extensions/writing-extensions.rst
@@ -229,7 +229,7 @@ The following TypeScript interface defines the contract for a service deployer:
         /**
         * In this phase, the service should remove all bindings on preDeploy resources.
         */
-        unBind?(ownServiceContext: ServiceContext<ServiceConfig>): Promise<IUnBindContext>;
+        unBind?(ownServiceContext: ServiceContext<ServiceConfig>, ownPreDeployContext: IPreDeployContext, dependentOfServiceContext: ServiceContext<ServiceConfig>, dependentOfPreDeployContext: IPreDeployContext): Promise<IUnBindContext>;
 
         /**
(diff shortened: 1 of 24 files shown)


dsw88:
Another option is to add methods to the contract for obtaining the context objects without actually running the phases themselves. So in the case of PreDeploy, for example, you'd have a method called `getPreDeployContext` that doesn't run PreDeploy, but is required to return a PreDeployContext with the information from the PreDeploy phase.

If we went that route, we'd want to add the following items to the phase provisioner contract:

* getPreDeployContext
* getDeployContext

Those items are currently the ones that return meaningful information in them.

I don't know that I like this too much better than just running PreDeploy, but it could be useful for other things. When I supported external dependencies early in the project it was implemented using methods like these in the contract that returned the deploy information without running the actual deploy itself.

dsw88:
After talking with @ThatJoeMoore I don't think we want to merge this PR. It seems better to implement this functionality by adding things like getPreDeployContext() and getDeployContext() to the contract. Those will not only serve for helping UnBind get what it needs, but they can also facilitate external references should we decide to implement them.

PR #555, opened 2018-09-19
Add polly support

### LINKED ISSUE #172: Polly - Add support for service
Polly is a text-to-speech HTTP service. This service should be fairly easy to add, it's mostly just wiring up the permissions to talk to the service.

mxnguyen350:
Resolves #172


mxnguyen350:
This isn't the final version, I just thought I would get your feedback

PR #562, opened 2018-09-24
Add comprehend

### LINKED ISSUE #558: Comprehend - Add support for service
This will likely go in the AI Services provisioner

mxnguyen350:
Resolves #558


handel/src/services/aiservices/comprehend-statements.json
```diff
@@ -0,0 +1,39 @@
+[
+    {
+        "Effect": "Allow",
+        "Action": [
+            "comprehend:BatchDetectDominantLanguage",
+            "comprehend:BatchDetectEntities",
+            "comprehend:BatchDetectKeyPhrases",
+            "comprehend:BatchDetectSentiment",
+            "comprehend:BatchDetectSyntax",
+            "comprehend:DescribeGominantLanguageDetectionJob",
```
dsw88: Spelling error here.

handel/src/services/aiservices/comprehend-statements.json
```diff
@@ -0,0 +1,39 @@
+[
+    {
+        "Effect": "Allow",
+        "Action": [
+            "comprehend:BatchDetectDominantLanguage",
```
dsw88: You can make the policy document shorter by using the `*` character. For example, if you want to grant permissions to all the API calls in Comprehend that start with `Describe`, then you can use a permission like `comprehend:Describe*`. This cuts down on the size of the policy, which is important if you have multiple services that you depend on.

So since we're granting permissions to pretty much everything in Comprehend, we can shorten it even further. If we're granting access to all the API calls, we can shorten it to a single action of just `comprehend:*`, which basically says "Allow access to all API calls on the comprehend service"

PR #570, opened 2018-10-08
Added application tags to deployment logs

yoshutch:
This will add the application logs in the handel file to the deployment log entry for each deployment/deletion.


handel/src/aws/dynamodb-calls.ts
```diff
@@ -120,6 +120,7 @@ export async function logHandelAction(lifecycleName: string, envResult: Environm
         EnvironmentName: envResult.environmentName,
         DeploymentStatus: envResult.status,
         DeploymentMessage: envResult.message,
+        ApplicationTags: handelFile.tags,
```
dsw88: The tags field in the Handel file is optional, so sometimes this will be undefined. It might be nice to have an empty object instead for the case where there are no application tags so that parsing it can be the same for consumers. So something like this:
```
ApplicationTags: handelFile.tags || {}
```

PR #571, opened 2018-10-08
Add support for non-NPM extensions

ThatJoeMoore:
Includes the following new Extension types:

* `file:` for project-specific extensions
* `github:`, `gitlab:`, and `bitbucket:` for SCM-as-a-service
* `git:` for generic Git repositories

This all just uses `npm install` under the hood, so the only real trick is the up-front parsing of the handelfile.


ThatJoeMoore:
Looks like I forgot to uncomment and fix some of my tests; lemme fix that before merging.

ThatJoeMoore:
Tests should be good to go now.

dsw88:
Looks awesome, thanks for adding this!

PR #576, opened 2018-10-26
changed password generators to make crypto random strings

### LINKED ISSUE #574: Convert password generation to use crypto-random-string
The following services will possibly need to be modified:

* Amazon MQ
* Aurora
* Aurora Serverless
* Elasticsearch
* MySQL
* Neptune
* PostgreSQL

mxnguyen350:
Resolves #574

### DIFF
--- a/handel/src/services/amazonmq/index.ts
+++ b/handel/src/services/amazonmq/index.ts
@@ -14,6 +14,7 @@
  * limitations under the License.
  *
  */
+import cryptoRandomString = require('crypto-random-string');
 import * as fs from 'fs';
 import {
     BindContext,
@@ -37,7 +38,6 @@ import {
     preDeployPhase,
     tagging
 } from 'handel-extension-support';
-import * as uuid from 'uuid/v4';
 import * as winston from 'winston';
 import { AmazonMQServiceConfig, HandlebarsAmazonMQTemplate } from './config-types';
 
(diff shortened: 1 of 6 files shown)


handel/src/services/amazonmq/index.ts
```diff
@@ -52,14 +52,14 @@ const AMAZON_MQ_PORTS = [
     61619 // WSS
 ];
 
-// This is pretty hacky...better add a good crypto random string soon
+// There are 36 charatcers because the uuid in the old code returned a 36 character username
```
dsw88: This service doesn't update anyway so the 36-character requirement doesn't need to be met. Let's remove these comments and make the username shorter; how about 12 characters? Let's leave the passwords at the longer length though.

handel/src/services/aurora/index.ts
```diff
@@ -202,7 +203,7 @@ export class Service implements ServiceDeployer {
         const stack = await awsCalls.cloudFormation.getStack(stackName);
         if (!stack) {
             const dbUsername = rdsDeployersCommon.getNewDbUsername();
-            const dbPassword = rdsDeployersCommon.getNewDbPassword();
```
dsw88: Let's keep using the rdsDeployersCommon instead of using the crypto random string directly. But we should modify the getNewDbUsername and getNewDbPassword functions in the RDS deployers common module to use the crypto random string generation.

Likewise let's make this change on the Aurora Serverless, MySQL, and PostgreSQL deployers.

mxnguyen350: When I looked into the code, I realized that the rdsDeployersCommon.getNewDbPassword already calls on the same crypto random string generator

dsw88: Oh cool, so really I guess we only need to change Amazon MQ then.

PR #589, opened 2018-12-05
fixed cloudwatch schema

### LINKED ISSUE #588: Event Pattern Used be a YAML syntax
https://github.com/byu-oit/handel/blob/dfa76dc89ec1adf6862e18fe21ab130ca5abfa02/handel/src/services/cloudwatchevent/params-schema.json#L21-L25

The Event Pattern section used to follow yaml syntax as an object. The schema is requiring a string yet the documentation and examples give a Yaml object

mxnguyen350:
resolves #588

### DIFF
--- a/handel/src/services/cloudwatchevent/params-schema.json
+++ b/handel/src/services/cloudwatchevent/params-schema.json
@@ -19,9 +19,13 @@
             "errorMessage": "Must be a string"
         },
         "event_pattern": {
-            "type": "string",
+            "type": "object",
             "description": "The list of event patterns on which to fire the event",
-            "errorMessage": "Must be a string"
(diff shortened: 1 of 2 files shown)


handel/src/services/cloudwatchevent/params-schema.json
```diff
@@ -19,9 +19,13 @@
             "errorMessage": "Must be a string"
         },
         "event_pattern": {
-            "type": "string",
+            "type": "object",
             "description": "The list of event patterns on which to fire the event",
-            "errorMessage": "Must be a string"
+            "items": {
```
dsw88: This needs to be a free-form object (i.e. type 'any')

PR #626, opened 2019-10-11
Feature permissions boundary

joshgubler:
If `permissions_boundary` is in the account config, Handel will add that permissions boundary to each role it creates.


ndpete:
do you want to add some documentation to the account config file? https://github.com/byu-oit/handel/blob/master/docs/handel-basics/account-config-file.rst

GaryGSC:
Seems like these doc updates haven't been published?
I don't seem them at https://handel.readthedocs.io/en/latest/handel-basics/account-config-file.html

Refs: https://github.com/byu-oit/handel/issues/630

